Issue #2, opened 2016-02-17
writeFile doesn't create a new file.


tamulj:
I'm a bit unclear why appendFile will add a new file but writeFile will throw an error instead of creating the file like the node fs will. Is this an oversight or intended behavior?

streamich:
Thanks, for pointing this out. It's an oversight, but now fixed:

    npm install memfs@0.0.8

PR #3, opened 2016-03-09
create file if not exist when writeFile

Samisdat:
https://github.com/streamich/memfs/issues/2

Just copied your code form appendFileSync to writeFileSync


streamich:
Thanks, fixed it

PR #6, opened 2016-12-15
Throw error with correct code when node exists

neezer:
I've the following helper method to create directories:

```js
const mkdirSync = fs => (dirPath) => {
  try {
    fs.mkdirSync(dirPath);
  } catch (e) {
    if (e.code !== 'EEXIST') {
      throw e;
    }
  }
};
```

Unfortunately, this doesn't work with `memfs`, as `addDir`/`addFile` will throw a non-system error if a node exists (with no code defined).

This PR throws an error closer to what Node itself would actually throw: https://nodejs.org/api/errors.html#errors_system_errors

---

Lemme know if you need me to change anything about this PR to get it merged in. Really enjoyed working with `memfs`, particularly for testing. Thanks for your hard work!

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,16 @@ var __extends = (this && this.__extends) || function (d, b) {
     function __() { this.constructor = d; }
     d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
+function SystemError(message, code) {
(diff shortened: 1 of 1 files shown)


neezer:
Aph, just realized your using TypeScript as your source... I'll re-open later once I update. Sorry!

PR #32, opened 2017-08-31
Mention that vol.toJSON will ignore empty folders

sapegin:


streamich:
`toJSON()` could also be extended to return folders, all hard links and symlinks.

PR #33, opened 2017-09-02
updated fs-monkey and fixed win32 issue

fearthecowboy:
**NOTE** This PR requires fs-monkey to have the [PR](https://github.com/streamich/fs-monkey/pull/2)  merged and published before this will work.

Using a memfs as for `node_modules` via fs-monkey was having a problem because unixify was inline-required (and gave problems during module resolution).

I inlined unixify in fs-monkey (as part of the [fix ](https://github.com/streamich/fs-monkey/pull/2) for win32 there and then pulled thru the implementation so that there was a few less dependencies (since fs-monkey was already a dep).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,12 +1,23 @@
 {
   "name": "memfs",
-  "version": "2.5.0",
+  "version": "2.5.2",
   "description": "In-memory file-system with Node's fs API.",
   "main": "lib/index.js",
   "keywords": [
-    "fs", "filesystem", "fs.js", "memory-fs", "memfs",
-    "file", "file system", "mount", "memory", "in-memory",
-    "virtual", "test", "testing", "mock"
+    "fs",
+    "filesystem",
+    "fs.js",
+    "memory-fs",
+    "memfs",
+    "file",
+    "file system",
+    "mount",
+    "memory",
+    "in-memory",
+    "virtual",
+    "test",
+    "testing",
+    "mock"
   ],
   "repository": {
     "type": "git",
@@ -14,8 +25,7 @@
   },
   "dependencies": {
     "fast-extend": "0.0.2",
-    "fs-monkey": "^0.1.3",
-    "unixify": "^1.0.0"
+    "fs-monkey": "https://github.com/fearthecowboy/fs-monkey/releases/download/0.1.5/fs-monkey-0.1.5.tgz"
   },
   "devDependencies": {
     "mocha": "3.4.2",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "fast-extend": "0.0.2",
-    "fs-monkey": "https://github.com/fearthecowboy/fs-monkey/releases/download/0.1.5/fs-monkey-0.1.5.tgz"
+    "fs-monkey": "~0.1.5"
   },
   "devDependencies": {
     "mocha": "3.4.2",
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -12,7 +12,7 @@ const util = require('util');
 
 
 const isWin = process.platform === 'win32';
-
+const unixify = isWin ? require("fs-monkey/lib/correctPath").unixify : p=>p;
 
 // ---------------------------------------- Types
 
@@ -373,9 +373,7 @@ export function pathToFilename(path: TFilePath): string {
 }
 
 function resolve(filename: string, base: string = process.cwd()): string {
-    filename = resolveCrossPlatform(base, filename);
-    if(isWin) filename = require('unixify')(filename);
-    return filename;
+    return unixify(resolveCrossPlatform(base, filename));
 }
 
 export function filenameToSteps(filename: string, base?: string): string[] {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,21 +1,31 @@
...
   "name": "memfs",
-  "version": "2.5.0",
+  "version": "2.5.2",
   "description": "In-memory file-system with Node's fs API.",
   "main": "lib/index.js",
   "keywords": [
-    "fs", "filesystem", "fs.js", "memory-fs", "memfs",
-    "file", "file system", "mount", "memory", "in-memory",
-    "virtual", "test", "testing", "mock"
+    "fs",
```
fearthecowboy: Sorry. npm v5 did this.

package.json
```diff
@@ -1,21 +1,31 @@
...
   ],
   "repository": {
     "type": "git",
     "url": "https://github.com/streamich/memfs.git"
   },
   "dependencies": {
     "fast-extend": "0.0.2",
-    "fs-monkey": "^0.1.3",
-    "unixify": "^1.0.0"
+    "fs-monkey": "~0.1.5"
```
fearthecowboy: REQUIRES [fix ](https://github.com/streamich/fs-monkey/pull/2) to be committed and published in fs-monkey first!

Oh, and I inlined unixify in fs-monkey and pulled it from that.

src/volume.ts
```diff
@@ -373,9 +373,7 @@ export function pathToFilename(path: TFilePath): string {
 }
 
 function resolve(filename: string, base: string = process.cwd()): string {
-    filename = resolveCrossPlatform(base, filename);
-    if(isWin) filename = require('unixify')(filename);
-    return filename;
+    return unixify(resolveCrossPlatform(base, filename));
```
fearthecowboy: `unixify` is an empty function if it's not win32, so no check at runtime needed.

streamich:
I don't have Windows machine unfortunately, but what is the root cause? `patchRequire` was not working because `memfs` uses `unixify`?

fearthecowboy:
The call for `require('unixify')` in memfs was first hit *after* the `patchRequire` was completed, so when it went to pull in `unixify`,  it couldn't find it because it wasn't in my volume that I had patched to.

So, it has to be required at module load time (so that it's not happening dynamically in the middle of a require elsewhere)

streamich:
So if we move this line

https://github.com/streamich/memfs/blob/master/src/volume.ts#L377

to `require` outside the `resolve` function, like this:

```js
const unixify = require('unixify');

function resolve(filename: string, base: string = process.cwd()): string {
    filename = resolveCrossPlatform(base, filename);
    if(isWin) filename = unixify(filename);
    return filename;
}
```

Would that work?

fearthecowboy:
That was my original fix -- I had to add `unixify` to `fs-monkey` (which is a dependency of memfs) and I didn't want to drag in so many tiny packages for fairly simple functionality. 

That fix is fine too; I was just trying to cut down on the extra stuff dragged in.

FYI, if you want to keep unixify as a dependency, you might want to do this instead

``` javascript
const unixify = isWin ? require("fs-monkey/lib/correctPath").unixify : p=>p;
```
That way you're not requiring `unixify` on non-windows at all.

streamich:
So, if we have `correctPath` in `memfs` this way, we don't need it in `fs-monkey`, as I understand it is basically problem with `memfs` because it requires `unixify` dynamically inside that `resolve` function?

If that is the case, I think it is better if we do something like this:

```ts
type TResolve = (filename: string, base?: string) => string;

let resolve: TResolve = (filename: string, base: string = process.cwd()) => {
    filename = resolveCrossPlatform(base, filename);
    if(isWin) filename = require('unixify')(filename);
    return filename;
}

if(isWin) {
    const {correctPath} = require('./correctPath');
    resolve = (filename, base) => correctPath(resolve(filename, path));
}
```

fearthecowboy:
Well, we still need `correctPath` in `fs-monkey` because a simple vol implementation in `fs-monkey` will still have paths that need to be fixed up there.

(the problems in `fs-monkey` were tangential to the `unixify` bug here)

streamich:
`memfs` internally represents paths in unix style, that is why it "normalizes" Windows paths using the `unixify` function.

Why does `fs-monkey` need to normalize the paths?

fearthecowboy:
Sorry for the delay -- was out sick.

using `fs-monkey`'s `patchRequire` is still probing for paths that are not normalized to unix-style paths.

Test case for stock monkey-fs that shows the problem:

```  javascript 

patchRequire = require ('fs-monkey').patchRequire;

let vol = {
    readFileSync: () => {},
    realpathSync: () => {},
    statSync: (path) => { console.log(path); }, // show the names being requested
};

patchRequire(vol);

require("/foo/bar");
```


output on windows:


```
C:\\fs-monkey-test> node .\test.js
\\?\C:\foo\bar
\\?\C:\foo\bar.js
\\?\C:\foo\bar.json
\\?\C:\foo\bar.node
module.js:472
    throw err;
    ^
```

with `patchRequire` , it's asking for NT pathnames, they are not getting normalized

PR #35, opened 2017-09-08
fixes accidental recursion with resolve function

fearthecowboy:
This fixes an accidental recursion (when the intent was to call the original function)


streamich:
Published under v2.5.4

PR #40, opened 2017-09-26
Make sure path is calculated correctly when renaming a file

skevy:
When calling `rename`, I noticed that sometimes (especially when dealing with paths at the volume root), the new path was being calculated improperly.

This fixes that, and I also added a regression test for `rename`.


streamich:
Thanks, not sure the fix is entirely correct though. It should fail for two level deep paths `/foo/bar`, I think.

Created a pull request that should fix this issue:

https://github.com/streamich/memfs/pull/42

PR #41, opened 2017-09-26
Make sure mode is OR'ed with the permission on init and when chmoding

skevy:
The `mode` variable inside a `Node` should not only indicate the type of Node it is (directory, file, etc), but should also contain information about it's permissions.

This PR makes sure that the mode is `|=`'ed with the `perm` variable when creating a new node and when chmoding. Also adds tests.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -38,6 +38,7 @@ var Node = (function (_super) {
         // Steps to another node, if this node is a symlink.
         _this.symlink = null;
         _this.perm = perm;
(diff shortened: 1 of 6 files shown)


streamich:
Permissions in `memfs` are actually not implemented yet. There are few bits here and there, but in general I would not rely on them right now too much.

Also, I don't think it will be possible in future to emulate all the permission features of a real file system.

streamich:
Released under `2.5.6` version.

PR #242, opened 2018-11-08
Housekeeping

streamich:
- Adds Prettier
- Adds TSLint
- Upgrades dependencies
- Adds `yarn.lock`
- Adds Husky and uses precommit and prepush hooks to run some linting stuff
- Upgrades TypeScript
- Adds `semantic-release` plugins to publish better info back to GitHub


streamich:
Made some general set-up improvements in this PR. Not sure why Jest now fails in Node 7 and 6. Will investigate that when have more time.

PR #264, opened 2018-11-29
chore(build): remove unnecessary copy of src/internal

### LINKED ISSUE #263: internal/error.rs causes typescript builds to break
Build script `tsc -p . && cpy src/internal lib/internal` copies also the error.ts file which subsequently gets then published causing the typescript projects to break.

Script could be enhanced with:
```
tsc -p . && cpy src/internal lib/internal && rm lib/internal/error.ts 
```

pizzafroide:
Fixes #263 

I am not sure why this copy was done, actually. I guess it is legacy code. @streamich, can you confirm that?

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   ],
   "scripts": {
     "clean": "rimraf lib types",
-    "build": "tsc -p . && cpy src/internal lib/internal",
+    "build": "tsc -p .",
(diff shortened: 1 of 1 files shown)


streamich:
Yes, its legacy, it used to be `.js` file copied from Node internals.

PR #277, opened 2018-12-11
Fixed constant "fd" value to start from 0x7fffffff instead of 0xffffffff

### LINKED ISSUE #274: File descriptor number is out of range
The following code throws `RangeError [ERR_INVALID_FD]: "fd" must be a positive integer`

```javascript
var memfs = require('memfs');
var tty = require('tty');

var fd = memfs.openSync('/test', 'w+');
var stream = new tty.WriteStream(fd);
```

This is because the generated file descriptor number exceeds the range of positive integers (32-bit) which causes it to be considered a negative integer. 

I believe the problem is in the Volume class, where the `static fd` variable is defined. It is initialized with 0xffffffff when it should be 0x7fffffff.

reinaldoarrosi:
Fixes #274

It's worth to mention that, even though this change does fix the error described in #274, the code in that issue will throw a different error because of the way that tty.WriteStream works internally.

Anyway, it's good to make sure that the file descriptors follow the unix guidelines.

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -535,7 +535,7 @@ export class Volume {
    * @type {number}
    * @todo This should not be static, right?
    */
-  static fd: number = 0xffffffff;
+  static fd: number = 0x7fffffff;
 
   // Constructor function used to create new nodes.
   // NodeClass: new (...args) => TNode = Node as new (...args) => TNode;


streamich:
Thanks!

streamich:
> ... will throw a different error because of the way that tty.WriteStream works internally.

What do you exactly mean? Why `tty.WriteStream` throws?

reinaldoarrosi:
tty.WriteStream does not work with arbitrary file descriptors, in fact even using `fs.openSync(<path_to_a_file>)`  to get a `fd` that points to an actual file and the using this `fd` to instantiate tty.WriteStream will fail with `EBADF: bad file descriptor, uv_tty_init`

I believe this happens because tty.WriteStream checks to see if the `fd` is one of the stdio (0 - stdin, 1 - stdout, 2 - stderr).

streamich:
Ok, thanks!

PR #308, opened 2019-01-27
feat(volume): add env variable to suppress fs.promise api warnings

ricsam:
To suppress experimental fs.promise api warnings the SUPRESS_EXPERIMENTAL_PROMISE_WARNINGS env variable can be set to a truthy value

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -139,3 +139,11 @@ you get from `require('fs')`. Here are some things this function does:
   ```
 
   - Adds constants `fs.constants`, `fs.F_OK`, etc.
+
+## Experimental fs.promise api warnings
+
+Supress warnings when using the promise api of fs by setting
+
+```js
+process.env['SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS'] = true;
+```
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -145,5 +145,5 @@ you get from `require('fs')`. Here are some things this function does:
 Supress warnings when using the promise api of fs by setting
 
 ```js
-process.env['SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS'] = true;
+process.env.MEMFS_DONT_WARN = true;
 ```
--- a/src/process.ts
+++ b/src/process.ts
@@ -1,11 +1,18 @@
 // Here we mock the global `process` variable in case we are not in Node's environment.
 
+export type SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS = 'SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS';
+export const SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS: SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS =
+  'SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS';
+
 export interface IProcess {
   getuid(): number;
   getgid(): number;
   cwd(): string;
   platform: string;
   nextTick: (callback: (...args) => void, ...args) => void;
+  env: {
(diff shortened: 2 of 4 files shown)


src/volume.ts
```diff
@@ -516,7 +516,7 @@ function validateGid(gid: number) {
 
 // ---------------------------------------- Volume
 
-let promisesWarn = true;
+let promisesWarn = process[SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS] ? false : true;
```
streamich: I think you missed `.env` here.

```js
let promisesWarn = !!process.env.MEMFS_DONT_WARN;
```

src/volume.ts
```diff
@@ -4,7 +4,7 @@ import Stats, { TStatNumber } from './Stats';
 import Dirent from './Dirent';
 import { Buffer } from 'buffer';
 import setImmediate from './setImmediate';
-import process from './process';
+import process, { SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS } from './process';
```
streamich: Can we treat this property as a string, so that we don't have to import it from the `./process`?

docs/reference.md
```diff
@@ -139,3 +139,11 @@ you get from `require('fs')`. Here are some things this function does:
   ```
 
   - Adds constants `fs.constants`, `fs.F_OK`, etc.
+
+## Experimental fs.promise api warnings
+
+Supress warnings when using the promise api of fs by setting
+
+```js
+process.env['SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS'] = true;
```
streamich:
```js
process.env.MEMFS_DONT_WARN = true;
```

src/process.ts
```diff
@@ -1,11 +1,18 @@
...
+  'SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS';
+
 export interface IProcess {
   getuid(): number;
   getgid(): number;
   cwd(): string;
   platform: string;
   nextTick: (callback: (...args) => void, ...args) => void;
+  env: {
+    [SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS]?: boolean;
```
streamich:
```js
MEMFS_DONT_WARN?: boolean
```

src/process.ts
```diff
@@ -1,11 +1,18 @@
 // Here we mock the global `process` variable in case we are not in Node's environment.
 
+export type SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS = 'SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS';
+export const SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS: SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS =
+  'SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS';
```
streamich: Do we really need to export a type and a string for this property. Can we just use the property directly everywhere, like

```js
process.env.MEMFS_DONT_WARN
```

src/volume.ts
ricsam: Great catch! :)

ricsam: I just noticed that this breaks the code. It should be `let promisesWarn = !process.env.MEMFS_DONT_WARN;`. I will submit another pr fixing this

Sleepful:
how do I use this?

I tried using `process.env.SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS = true`
but warning is still displayed and when looking at `process.env` I don't see the `SUPPRESS_EXPERIMENTAL_PROMISE_WARNINGS `

streamich:
Try

```js
process.env.MEMFS_DONT_WAR = true;
```

PR #373, opened 2019-05-01
fix: 🐛 accept `null` as value in `fromJSON` functions

G-Rath:
I created the `DirectoryJSON` alias just to make things easier.

I looked at enabled `strictNullChecks`, but this package just wasn't designed for it :joy:
Ideally, it'd be good to have it on at some point, or otherwise ensure all methods that can return `null` or `undefined` as thusly typed.

I've not linked the issue with this PR, as the issue could be used to track the above, and this is the change that I need ASAP to continue work on my current project.

Once I have some time, I'll be happy to look into making this library more `strictNullChecks` compatible.

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -808,7 +808,7 @@ export class Volume {
     });
   }
 
-  private _toJSON(link = this.root, json = {}, path?: string) {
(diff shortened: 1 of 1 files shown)


streamich:
Sorry for the delay, I somehow missed notification about this PR.

PR #378, opened 2019-05-18
feat: allow exporting file paths in Volume.toJSON

saboya:
This allows files path to be specified in `Volume.toJSON`. Without this patch, files specificed would show up in the resulting `JSON`, but with its contets set to `null`.

Test is included.


saboya:
@streamich Any thoughts on this?

PR #409, opened 2019-07-13
Make strict

### LINKED ISSUE #372: Not compatible with strictNullChecks
The feature implemented by #107 is not compatible when `strictNullChecks` is `true`, as it's typings only allow `string`.

ss+(2019-04-30+at+04 31 37)

I think the right typing would be `string | null`:

```
    fromJSON(json: {
        [filename: string]: string | null;
    }, cwd?: string): void;
```

----

On an aside, I'd love to contribute to this project - I've started using it for my test fixture, and it's working great. I think there's a couple of places that could be improved to make it gel nicer, such as `toJSON`, which I think could benefit greatly from supporting nesting ala `mock-js`.

If you're interested in having a chat, flick me an email :)

(and if I get the time, I'll make a PR for this myself)

G-Rath:
Finally got around to finishing it off :D

fixes #372 

I'd be interested in doing `noImplicitAny` next, and then trying eliminate all the casts.
Having worked w/ pretty much the entire codebase now, I've got a few ideas for improvements (mostly "cleanup/refactor" stuff), such as:

* Looking to split up `volume` (its size means it takes a bit longer to parse)
* Expanding the test suite to ensure stricter 1:1 w/ native fs (off the top of my head: a number of methods don't throw the correct parameter-type errors)
* Refining types to ensure stricter 1:1 w/ native fs (primarily ensuring type converts are supported i.e string to number)
* Improving overloads so that they're less hacky
* making the code more consistent overall

But it's your library, and I don't want to step on toes (or make it sound like you've done a soddy job or anything - quite the opposite; I love this library), so let me know what you're open to 😄 

No one seems to be crying out for this, so I think it'd be fine to wait for some of the above possible changes to be done before releasing this as a major (if you're interested in more work being done).


streamich:
> But it's your library, and I don't want to step on toes (or make it sound like you've done a soddy job or anything - quite the opposite; I love this library), so let me know what you're open to 😄

@G-Rath I'm open to all of those suggestions, I think all of those are good. Especially making this library more in line with Node, I initially made the errors same as what Node was throwing, but as Node develops it changes over time and errors get out of sync.

G-Rath:
@streamich awesome! I'll look at start writing some issues up (for my own tracking more than anything) & start chipping away at things when I have the time.

For my own sanity I'll probably try to push things through via smallish PRs; but first I'll just quickly fix up the :Possibly undefined"s I somehow missed O.o

src/volume.ts
```diff
@@ -1099,14 +1100,14 @@ export class Volume {
...
-    let position: number;
+    let encoding: TEncoding | undefined;
+    let offset: number | undefined;
+    let length: number | undefined;
+    let position: number | undefined;
 
     const isBuffer = typeof a !== 'string';
     if (isBuffer) {
-      offset = b | 0;
+      offset = (b || 0) | 0;
```
streamich: Is `|| 0` necessary?

src/volume.ts
```diff
@@ -636,30 +618,40 @@ export class Volume {
...
-      link.vol = null;
-      link.parent = null;
+
       return true;
     }
     return false;
   }
 
   private newInoNumber(): number {
-    if (this.releasedInos.length) return this.releasedInos.pop();
```
streamich: Wouldn't checking for `.length` be faster?

src/volume.ts
```diff
@@ -1306,7 +1309,7 @@ export class Volume {
     const srcFilename = pathToFilename(src);
     const destFilename = pathToFilename(dest);
 
-    return this.copyFileBase(srcFilename, destFilename, flags | 0);
+    return this.copyFileBase(srcFilename, destFilename, (flags || 0) | 0);
```
streamich: Is `|| 0` needed?

src/volume.ts
```diff
@@ -1400,22 +1403,23 @@ export class Volume {
 
   symlink(target: TFilePath, path: TFilePath, callback: TCallback<void>);
   symlink(target: TFilePath, path: TFilePath, type: TSymlinkType, callback: TCallback<void>);
-  symlink(target: TFilePath, path: TFilePath, a, b?) {
-    const [type, callback] = getArgAndCb<TSymlinkType, TCallback<void>>(a, b);
+  symlink(target: TFilePath, path: TFilePath, a: TSymlinkType | TCallback<void>, b?: TCallback<void>) {
+    const callback: TCallback<void> = validateCallback(typeof a === 'function' ? a : b);
```
streamich: Nice!

src/volume.ts
```diff
@@ -1621,16 +1626,22 @@ export class Volume {
...
       const list: Dirent[] = [];
       for (const name in link.children) {
-        list.push(Dirent.build(link.children[name], options.encoding));
+        const child = link.getChild(name);
+
+        if (!child) {
+          continue;
+        }
+
+        list.push(Dirent.build(child, options.encoding));
```
streamich: I'm thinking `child` should always be truthy here, maybe use non-null assertion instead `child!`.

src/volume.ts
```diff
@@ -1747,9 +1760,11 @@ export class Volume {
   truncate(id: TFileId, callback: TCallback<void>);
   truncate(id: TFileId, len: number, callback: TCallback<void>);
   truncate(id: TFileId, a: TCallback<void> | number, b?: TCallback<void>) {
-    if (isFd(id)) return this.ftruncate(id as number, a as any, b);
+    const len: number = typeof a === 'number' ? a : 0;
+    const callback: TCallback<void> = validateCallback(typeof a === 'number' ? b : a);
+
+    if (isFd(id)) return this.ftruncate(id as number, len, callback);
```
streamich: Maybe we can make `isFd` a guard function, then `as number` will not be needed.

```ts
const isFd: (fd: number): fd is number => ...
```

src/volume.ts
```diff
@@ -1836,8 +1851,9 @@ export class Volume {
   mkdir(path: TFilePath, callback: TCallback<void>);
   mkdir(path: TFilePath, mode: TMode | IMkdirOptions, callback: TCallback<void>);
   mkdir(path: TFilePath, a: TCallback<void> | TMode | IMkdirOptions, b?: TCallback<void>) {
-    const [options, callback] = getArgAndCb<TMode | IMkdirOptions, void>(a, b);
-    const opts = getMkdirOptions(options);
+    const opts: TMode | IMkdirOptions = getMkdirOptions(a);
```
streamich: Here `a` could be a callback function, but we don't want to call `getMkdirOptions` with a function.

src/__tests__/volume.test.ts
```diff
@@ -416,7 +412,7 @@ describe('volume', () => {
       it('Closes file without errors', done => {
         vol.open('/test.txt', 'w', (err, fd) => {
           expect(err).toBe(null);
-          vol.close(fd, err => {
+          vol.close(fd || -1, err => {
```
streamich: Why `-1`?

src/volume.ts
```diff
@@ -636,30 +618,40 @@ export class Volume {
   createLink(): Link;
   createLink(parent: Link, name: string, isDirectory?: boolean, perm?: number): Link;
   createLink(parent?: Link, name?: string, isDirectory: boolean = false, perm?: number): Link {
-    return parent ? parent.createChild(name, this.createNode(isDirectory, perm)) : new this.props.Link(this, null, '');
+    if (!parent) {
+      return new this.props.Link(this, null, '');
+    }
+
+    if (!name) {
+      throw new Error('createLink: name cannot be empty');
```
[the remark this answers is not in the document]
G-Rath: This is b/c `name` can technically be undefined - I looked around, tested, etc and as far as I could see it never actually is.

Hence I added this w/ the idea of if it does throw, someone will make an issue and it'll be pretty clear we have to account for this.

> Maybe functions that call createLink should make sure name is always truthy instead?

That would be the ideal thing to do - ultimately, the smaller the overloads we have, the better/easier things will be (as TS gets super strict when they come into play, due to the dynamic nature of JS)

I'll have a look into this, but pretty sure I went w/ this b/c I don't understand what it's doing relative to Node well enough to know what to use for `name`? 

> Node does not throw such an error.

Where in Node is this? Node doesn't have a `createLink` function; if you could point me at an example of the desired behaviour, I should be able to ensure best implementation (as that was what I couldn't find the first time :joy:)

src/volume.ts
G-Rath: Yes, b/c of how overloading & type narrowing works:

TS doesn't change the type of things indirectly, b/c it's a massive can of worms.  This means you have to do some redundant checks when working w/ method overloads.

src/volume.ts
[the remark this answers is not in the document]
streamich: > Hence I added this w/ the idea of if it does throw, someone will make an issue...

That's good.

> Where in Node is this? Node doesn't have a createLink function ...

Yeah, that was my point, that Node does not throw something like this, so our users should not receive it either. But I like the idea that they should never get this error, and if they do they can create an issue.

src/volume.ts
G-Rath: Sure.

Personally I'm not a fan of using `!` if I can avoid it, as it means you're putting the trust completely on all the code above the line (and so if it changes, TS won't complain), but your call :)

src/volume.ts
G-Rath: I actually talked about this w/ some mates on the TS gitter channel.

The problem is that `isFd` is not just checking if its a number, but also that it's greater than 0.

This is a problem b/c:

```
const guard = (i: string | number): i is number => typeof i === 'number' && i > 0;

const fn = (v: string | number): string => {
  if(guard(v)) {
    // v is string | number, & "is number" === true, so it must be a number!
    return v.toFixed().toString();
  }

 // v is string | number, & "is number" !== true, so it must be a string!
  return v.toUpperCase();
}

fn(1);
fn(2);
fn('hello world');
fn(-1); // error, no such method "toUpperCase" on number!
```

src/__tests__/volume.test.ts
G-Rath: b/c it could be undefined, and `-1` is typically an error code :P

It's purely arbitrary - it's in a test file, and not what the test is testing, so I just picked a sane common `|| <fallback>` number value.

I'm happy to change it to anything - the important thing is it needs to be `|| <number>` for TS to be happy.

(In saying that, it's a file descriptor, which are not allowed to be negatives right...?)

G-Rath:
@streamich 

> It is up to you, I can create next branch where you can merge stuff for the next major; or we can release this PR as a breaking change.

I think the `next` branch is the way to go (at least for now) - that way we can chip away at improvements, and if someone asks for that stuff to be released we can just publish what we've got :)

Plus it might be cool to do some `rc` releases every couple of big PRs, like this one for incremental migration.

I think I'll try to pull a few small things into master just to make things consistent - one of the big ones that spring to mind is switching to `throw createError`, just b/c it changes so many lines w/o changing any runtime behaviour.

src/__tests__/volume.test.ts
streamich: -1 is good

streamich:
Created `next` branch.

src/volume.ts
G-Rath: Yes, as it's technically possible for `flags` to be undefined.

G-Rath:
Just realised my tests are breaking b/c I'm out of sync w/ master due to

```
let p: IProcess = typeof process !== 'undefined' && process;
```

Not entirely sure why github doesn't let me update my branch here - could be b/c it's a fork.
I'll try to update my fork - apologies in advance if a weird merge commit appears 😬

G-Rath:
I've made #410 to fix the new process code.

Also, I've got to ask you about `internal/events.ts` at some point :joy:

On an aside, I probably won't have time to start on any of the major things until at least the weekend, but will probably have the time to do some minor stuff, like package updates for a couple of major versions.

streamich:
Thanks for this great work, I'm gonna merge it here: https://github.com/streamich/memfs/pull/470

Issue #419, opened 2019-08-19
Better support for older Node.js versions


rangoo94:
Hello,

I came to issue that right now `memfs` is not working on older Node.js version than 5.10, while it can be easily fixed.

* `Buffer.from` and `Buffer.allocUnsafe` - these methods have been [added](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array) in Node.js 5.10
  * [`buffer-from`](https://www.npmjs.com/package/buffer-from) and [`buffer-alloc-unsafe`](https://www.npmjs.com/package/buffer-alloc-unsafe) ponyfills can be used

streamich:
Thanks, I like the idea of supporting older versions of `Buffer`, but it somehow feels that adding two extra dependencies for everyone for something like this is too much.

Maybe users should polyfill this themselves? Or we can polyfill this with one line, like this:

```js
if (!Buffer.from) Buffer.from = Buffer.allocUnsafe = (...args) => new Buffer(...args);
```

PR #421, opened 2019-08-19
fix(#419) add support for Node.js 0.11+

rangoo94:
Use ponyfills for `Buffer.allocUnsafe` and `Buffer.from` to get better support (#419).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,8 @@
     "url": "https://github.com/streamich/memfs.git"
   },
   "dependencies": {
+    "buffer-alloc-unsafe": "1.1.0",
+    "buffer-from": "1.1.1",
     "fast-extend": "0.0.2",
     "fs-monkey": "^0.3.3"
   },
@@ -33,6 +35,7 @@
     "@semantic-release/changelog": "3.0.4",
     "@semantic-release/git": "7.0.16",
     "@semantic-release/npm": "5.1.13",
+    "@types/buffer-from": "1.1.0",
     "@types/jest": "23.3.14",
     "@types/node": "10.14.15",
     "cpy-cli": "2.0.0",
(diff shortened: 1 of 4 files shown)


streamich:
Thanks, I like the idea of supporting older versions of `Buffer`, but it somehow feels that adding two extra dependencies for everyone for something like this is too much.

Maybe users should polyfill this themselves? Or we can polyfill this with one line, like this:

```js
if (!Buffer.from) Buffer.from = Buffer.allocUnsafe = (...args) => new Buffer(...args);
```

rangoo94:
Hi, thanks for your response.

I would strongly recommend to **not** attach any polyfills with library, as they may affect other packages. There should be either used ponyfill, or it should be left for users themselves.

If users will have to deal with it on their side, they still need to attach `Buffer` polyfills, so they will affect environment anyway (what may make some issues).

Ponyfills which I have attached here are pretty small (~1kb of code, ~15kb of package), while gives the proper interface for these operations ([alloc-unsafe](https://github.com/LinusU/buffer-alloc-unsafe/blob/master/index.js) and [from](https://github.com/LinusU/buffer-from/blob/master/index.js)).

If you would really like to simplify these methods without external packages, you could use something like:

```ts
// bufferUtils.ts

function ownBufferCreate (size: number): Buffer {
  if (typeof size !== 'number') {
    throw new Error("Invalid buffer size");
  }

  // [...] Other validation

  return new Buffer(size);
}

function ownBufferFrom (input: any): Buffer {
  // [...] implementation
}

export const create = (typeof Buffer !== 'undefined' && Buffer.allocUnsafe) || ownBufferCreate;
export const from = (typeof Buffer !== 'undefined' && Buffer.from) || ownBufferFrom;


// usageExample.ts

import * as buffer from './bufferUtils';

// [...]

buffer.from(arr);
buffer.create(10);
```

streamich:
Can those two functions be implemented like this?

```ts
const bufferFrom = Buffer.from || (...args) => new Buffer(...args);
const bufferAllocUnsafe = Buffer.allocUnsafe || (...args) => new Buffer(...args);
```

It seems to be way less code, and do everything we need here.

PR #436, opened 2019-10-01
Omit promises warning before actual use

jonathantneal:
This change allows the creation of new volumes without emitting a promises warning. The warning is still omitted when the `promises` property is accessed for the first time on the volume itself, as expected.

---

This **should** emit a warning:

```js

const fs = new Volume();

console.log(fs.promises);
```

This **should not** emit a warning:

```js

const fs = new Volume();
```

---

**About the "true" string**:

```js
process.env.MEMFS_DONT_WARN = 'true';
```

The string `"true"` is used over the boolean `true` in order to respect the typings of `process.env`.


streamich:
I'm thinking we can remove this warning altogether.

streamich:
Removing warnings here: https://github.com/streamich/memfs/pull/445

PR #463, opened 2019-11-16
Fix statBase, realpathBase and getChild

JIACHENG9:
Why this pr:

1. statBase, realpathBase

`/a/b/index.js` if b is a symlink, stat, realpath function will fail

2. getChild

`/a/toString` if path includes `toString`, it will fail

I fork this repo and publish `memfs-nodebowl` for https://github.com/nodebowl/nodebowl, when this pr merge and publish, I will change it to `memfs`.

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -301,7 +301,9 @@ export class Link extends EventEmitter {
   }
 
   getChild(name: string): Link {
-    return this.children[name];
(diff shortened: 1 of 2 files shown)


streamich:
Thanks! I'll merge it here: https://github.com/streamich/memfs/pull/463

PR #485, opened 2019-12-24
fix(watch): trigger change event on a folder when children are added/deleted

imsnif:
Hey there,

In node's `fs`, when doing `fs.watch` on a folder, the listener is triggered when files are added or deleted from that folder. This PR mirrors this behaviour for memfs as well.

I did not add tests because I could not find any tests for `watch`. I'd be happy to write some, but feel this might be out of scope for this PR? Let me know what you think.

Thanks!


streamich:
Tests are always welcome, but I'm happy to accept a good change even without tests.

PR #486, opened 2019-12-24
fix(rmdir): support recursive option

imsnif:
Hey there,

So another PR for today :)
This one adds the `recursive` capabilities of `rmdir` and `rmdirSync` https://nodejs.org/api/fs.html#fs_fs_rmdir_path_options_callback

This one does include tests, but only for rmdirSync (as I could not find base tests for `rmdir`). As in the previous PR, I'd be happy to add them in a separate PR if desired.

Thanks again!

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -342,6 +342,17 @@ const getMkdirOptions = (options): IMkdirOptions => {
   return extend({}, mkdirDefaults, options);
 };
 
+// Options for `fs.rmdir` and `fs.rmdirSync`
+export interface IRmdirOptions {
+  recursive?: boolean;
+}
+const rmdirDefaults: IRmdirOptions = {
+  recursive: false,
+};
+const getRmdirOptions = (options): IRmdirOptions => {
+  return extend({}, rmdirDefaults, options);
+};
+
 // Options for `fs.readdir` and `fs.readdirSync`
 export interface IReaddirOptions extends IOptions {
   withFileTypes?: boolean;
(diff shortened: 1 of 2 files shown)


src/volume.ts
```diff
@@ -1901,20 +1912,25 @@ export class Volume {
...
+    this.rmdirBase(pathToFilename(path), options);
   }
 
-  rmdir(path: TFilePath, callback: TCallback<void>) {
+  rmdir(path: TFilePath, callback: TCallback<void>);
+  rmdir(path: TFilePath, options: IRmdirOptions, callback: TCallback<void>);
+  rmdir(path: TFilePath, a: TCallback<void> | IRmdirOptions, b?: TCallback<void>) {
+    const opts: IRmdirOptions = getRmdirOptions(a);
+    const callback: TCallback<void> = validateCallback(typeof a === 'function' ? a : b);
     this.wrapAsync(this.rmdirBase, [pathToFilename(path)], callback);
```
streamich: This is probably missing `opts`

```js
this.wrapAsync(this.rmdirBase, [pathToFilename(path), opts], callback);
```

Maybe worth adding a test for this, something like

```js
      it('Async remove dir /dir1/dir2/dir3 recursively', done => {
        const vol = new Volume();
        vol.mkdirSync('/dir1/dir2/dir3', { recursive: true });
        vol.rmdir('/dir1', { recursive: true }, () => {
          expect(!!vol.root.getChild('/dir1')).toBe(false);
          done();
        });
      });
```

imsnif:
*slaps forehead* good catch! Thanks :)

I adjusted the test to get "dir" rather than "/dir". This was also a mistake in my original test, so fixed that as well.

PR #487, opened 2019-12-25
fix(watch): suppress event-emitter warnings

imsnif:
Hello again :)

So, before this patch - if we were watching more than 11 files in the same folder, we'd get warnings like these:
```
(node:13033) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 child:delete listeners added to [Link]. Use emitter.setMaxListeners() to increase limit
```

This was because the `FSWatcher` would listen to the parent delete event to know when to emit a `rename` event for this node. Since in this case this is clearly not a memory leak, I think it makes sense to just increase the limit by 1 before adding the watcher. Let me know if you think differently and I'll try to find a different solution.

As mentioned in https://github.com/streamich/memfs/pull/485, I'd be happy to write some tests for `watch` that will include testing this functionality.

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -2566,6 +2566,7 @@ export class FSWatcher extends EventEmitter {
     const parent = this._link.parent;
     if (parent) {
       // parent.on('child:add', this._onParentChild);
+      parent.setMaxListeners(parent.getMaxListeners() + 1);
       parent.on('child:delete', this._onParentChild);
     }
 


streamich:
@imsnif Do you know how this works in real `fs` module? Would it show the warning? Or suppress it somehow?

imsnif:
My local tests (running webpack-dev-server) show that this doesn't happen with the real fs module in the same situation, but I'm afraid at the moment it would not be easy for me to provide a reproducible real-world scenario. :)

I looked a little bit through node's fs code, and if I'm not mistaken they handle this a little differently (on the C++ side of things): https://github.com/nodejs/node/blob/7e5bf80fd1a06655564ea88f5d7867b3f4fc7da0/src/fs_event_wrap.cc#L198

streamich:
Thanks, for good work, I've invited you as [collab](https://github.com/streamich/memfs/invitations).

imsnif:
Thanks for the invitation @streamich ! We're relying on this lib quite a bit - so I'll consult you if we want to make any big changes.

streamich:
@imsnif It will automatically release if you merge to `master`. But please wait for a review at least 3 days before merging without a review, unless it is a hotfix.

imsnif:
@streamich - will do.

PR #502, opened 2020-02-15
Remove dom requirement

G-Rath:


src/volume.ts
```diff
@@ -143,7 +143,7 @@ function formatError(errorCode: string, func = '', path = '', path2 = '') {
     case ENOSYS:
       return `ENOSYS: function not implemented, ${func}${pathFormatted}`;
     default:
-      return `${errorCode}: error occurred, ${func}${pathFormatted}`;
+      return `${errorCode} : error occurred, ${func}${pathFormatted}`;
```
streamich: Is this behaviour the same in Node.js? I.e. does it not add a space here? All the other error cases above don't have the space.

G-Rath: Oh crumbs, that wasn't meant to be committed 😬

That was b/c Grazie currently has a bug that causes it to throw, so I put it there to stop it while I worked, but then had to rebase a bunch.

Good catch!

G-Rath:
```suggestion
      return `${errorCode}: error occurred, ${func}${pathFormatted}`;
```

src/volume-localstorage.ts
```diff
@@ -29,7 +29,7 @@ export class ObjectStore {
   }
 }
 
-export function createVolume(namespace: string, LS: Storage | object = localStorage): new (...args) => Volume {
+export function createVolume(namespace: string, LS: IStore | object = localStorage): new (...args) => Volume {
```
G-Rath: This means we can't drop `dom` from lib, but is fine b/c it'll only come into play if this file is imported, which makes sense: don't import `volume-localstorage` if `localStorage` isn't defined 😄

streamich: That's fine, this file is not used anyways, there was an attempt to make `memfs` persist in `localStorage`, but it was never finished.

PR #503, opened 2020-02-15
Update jest

G-Rath:
Again did them all together, and also switched test env to `node` for extra speed.

Also chucked in a type for a private property on `Volume`, as I was hanging around there tracking down the memory leak: it seems that calling `unwatchFile` from within the `watchFile` callback doesn't get applied completely until the next tick.

This could technically be a bug, but I don't think it's an issue outside of situations like with jest since this is happening because jest considers anything on the next tick after `done` is called a leak.

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -1020,8 +1020,10 @@ describe('volume', () => {
         vol.writeFileSync('/lol.txt', '1');
         setTimeout(() => {
           vol.watchFile('/lol.txt', { interval: 1 }, (curr, prev) => {
-            vol.unwatchFile('/lol.txt');
-            done();
+            process.nextTick(() => {
+              vol.unwatchFile('/lol.txt');
+              done();
+            });
           });
           vol.writeFileSync('/lol.txt', '2');
         }, 1);
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
       "js",
       "jsx"
     ],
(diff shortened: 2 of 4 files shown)


G-Rath:
This is failing due to `jest-snapshot` relying on [a version of `semver` that doesn't support node 8](https://github.com/facebook/jest/issues/9450).

It's a bug that's been fixed, but who knows when it'll be released - for now I'm going to leave this open and I'll update it when the next version of jest is released.

G-Rath:
This is now failing because `ts-jest` updated `mkdirp` to v1, which broke node@8 support.

Have opened an [issue](https://github.com/kulshekhar/ts-jest/issues/1550) about getting it addressed.

src/__tests__/volume.test.ts
```diff
@@ -1020,8 +1020,10 @@ describe('volume', () => {
         vol.writeFileSync('/lol.txt', '1');
         setTimeout(() => {
           vol.watchFile('/lol.txt', { interval: 1 }, (curr, prev) => {
-            vol.unwatchFile('/lol.txt');
-            done();
+            process.nextTick(() => {
```
jayaddison, line 1023: This may be ancient history at this point @G-Rath, so no worries if it's beyond recall; I was wondering if you managed to figure out what was going on with the watch/unwatch behaviour here, and whether there was anything reported upstream for it?  Cheers!

PR #506, opened 2020-02-16
Use object assign instead of fast extend

G-Rath:
Because I like looking at really tiny numbers that don't matter in the real world, I did a micro benchmark of `fast-extend` vs `Object.assign` & spread:

```
extend 0.0065383616383615385
assign 0.003307592407592418
spread 0.003464735264735254
```

`Object.assign` consistently performs hits 0.003 while extend only ever gets as good as 0.004 every so often - it's most consistently at 0.005 if not higher.

However, `Object.assign` is Node 4 & above, and while `travis` tests against 8+, the engines field isn't in the `package.json`.

Hence, I've set the engine field in this PR: I think it's fine to not make this a major release, since 8 itself is EOL, and previous conversations have mentioned that we're only wanting to support the latest version of FS as it's otherwise too much overhead.

This should also mean that renovate will stop hitting us with PRs for majors that remove support for node8 🎉 

<details>
<summary>code used for performance test</summary>

```
const { PerformanceObserver, performance } = require('perf_hooks');

const { extend } = require('fast-extend');
const times = {
  extend: [],
  assign: [],
  spread: [],
};
const obs = new PerformanceObserver(list => {
  const entries = list.getEntries();

  entries.forEach(entry => times[entry.name].push(entry.duration));
});

obs.observe({ entryTypes: ['function', 'measure'] });

const loops = 1000;

const b = { bigint: true, p: {} };
const c = { bigint: false, p: { mine: 'hello' }, c: {}, def: {} };

for (let i = 0; i < loops; i++) {
  performance.mark(`extend:${i}:start`);
  extend({}, b, c);
  performance.mark(`extend:${i}:stop`);
}

for (let i = 0; i < loops; i++) {
  performance.mark(`assign:${i}:start`);
  Object.assign({}, b, c);
  performance.mark(`assign:${i}:stop`);
}

for (let i = 0; i < loops; i++) {
  performance.mark(`spread:${i}:start`);
  ({ ...b, ...c });
  performance.mark(`spread:${i}:stop`);
}

for (let i = 0; i < loops; i++) {
  Object.keys(times).forEach(name => performance.measure(name, `${name}:${i}:start`, `${name}:${i}:stop`));
}

Object.keys(times).forEach(name => {
  const durations = times[name];

  console.log(name, durations.reduce((sum, dur) => sum + dur, 0) / (durations.length + 1));
});

obs.disconnect();
```
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,9 @@
     },
     "testRegex": ".*/__tests__/.*\\.(test|spec)\\.(jsx?|tsx?)$"
   },
+  "engines": {
+    "node": ">= 8.3.0"
+  },
   "release": {
     "verifyConditions": [
       "@semantic-release/changelog",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "testRegex": ".*/__tests__/.*\\.(test|spec)\\.(jsx?|tsx?)$"
   },
   "engines": {
-    "node": ">= 8.3.0"
+    "node": ">= 4.0.0"
   },
   "release": {
     "verifyConditions": [
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "url": "https://github.com/streamich/memfs.git"
   },
   "dependencies": {
-    "fast-extend": "1.0.2",
     "fs-monkey": "0.3.3"
   },
   "devDependencies": {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "lib": ["es5", "dom", "es2015.promise"],
+    "lib": ["ES2017", "dom"],
     "module": "commonjs",
     "removeComments": false,
     "noImplicitAny": false,
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -12,7 +12,6 @@ import { constants } from './constants';
 import { EventEmitter } from 'events';
 import { TEncodingExtended, TDataOut, assertEncoding, strToEncoding, ENCODING_UTF8 } from './encoding';
 import * as errors from './internal/errors';
-const { extend } = require('fast-extend');
 import util = require('util');
 import createPromisesApi from './promises';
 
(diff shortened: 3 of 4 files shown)


G-Rath:
I believe `yarn` enforces the engine field while npm only warns - there is a flag to have it ignore it, but I'm happy to drop the engines field completely if you are :)

streamich:
Besides Renovate are there advantages of enforcing Node version? I would like users who use Node 6 or 8.2 still be able to use this package, without unnecessary complexities like setting a CLI flag. If there are no major benefits, I would suggest to not enforce min Node version. Unless `memfs` actually does not work in Node 6.

G-Rath:
The primary advantage with supporting node 8+ is native `await/async` (specifically, node 7.6+) - transpiling that makes up ~10% of the package size. This is because we're targeting ES5 which means generators are also transpiled.

Targeting ES2017 means we can use native `await/async` & generators, removing the biggest of the transpile costs, as well as a few others such as arrow functions & `__assign`.

I don't have any strong feelings aside from not wanting to get weighed down by having to strongly support such old versions of node - it would be ideal to target node@8 since that version has the bulk of the ES6 features everyone knows and loves, but can also target node@6 if you're set on it.

Overall, Node itself is fairly stable so really there shouldn't be a reason to not update to latest, but 🤷‍♂

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "lib": ["es5", "dom", "es2015.promise"],
```
streamich: I've missed that before, but now that you mentioned it in discussion, if we change build target to `es2017` then for browser users `memfs` will not work in older browsers. I think instead of changing the default build we can add another build target, which targets `es2017`, for example:

```ts
import { Volume } from 'memfs/es`;
```

streamich:
> Targeting ES2017 means we can use native await/async & generators...

We still need to transpile to ES5 for browser users. Maybe we can have two builds?

```ts
import { Volume } from 'memfs';
import { Volume } from 'memfs/es';
```

G-Rath:
> We still need to transpile to ES5 for browser users. Maybe we can have two builds?

I don't think it's worth having two builds.

More than happy to pin the engines property to `4` and call it a day :)

package.json
```diff
@@ -62,6 +61,9 @@
     },
     "testRegex": ".*/__tests__/.*\\.(test|spec)\\.(jsx?|tsx?)$"
   },
+  "engines": {
+    "node": ">= 8.3.0"
```
G-Rath:
```suggestion
    "node": ">= 4.0.0"
```

PR #573, opened 2020-09-15
Add argument validation for readBase and writeBase

Meigyoku-Thmn:
These 2 functions don't check the input at all (NodeJS does). So I make a validating function for them.


G-Rath:
Closing due to age

PR #592, opened 2020-10-27
fix: change to use caret for fs-monkey

### LINKED ISSUE #591: [Bug] should fs-monkey be ^1.0.1 instead of 1.0.1?
Currently fs-monkey is "1.0.1". Is it expected? Otherwise, it will have two versions running

wood1986:
fix #591

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "url": "https://github.com/streamich/memfs.git"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


G-Rath:
I'm going to close this because of the reasons outlined in my comment [here](https://github.com/streamich/memfs/issues/591#issuecomment-718213399); namely that Renovate will just open a PR reverting this if its merged.

Issue #533, opened 2020-05-06
Support for case sensitivity settings


piotr-oles:
Hi!
As you probably know, there are 2 options for case sensitivity: 
 * Windows use case sensitive paths, 
 * other systems like macOS or Linux use case insensitive paths

I checked how `memfs` behave and it seems that it uses case sensitive approach:
```javascript
const { fs } = require('memfs');

fs.writeFileSync('/my/Case/sensitive/Path.txt', 'content');

console.log(fs.readFileSync('/my/case/sensitive/path.txt'));
// logs undefined
// if I would use native fs on macOS or Linux, I would get "content" logged
// so it's kind of inconsistency between memfs and fs
```
It would be nice if:
 1. there would be an option to configure it
 2. the default behavior would be to use system settings

streamich:
> 1. there would be an option to configure it

Yes, it seems we could add support for case insensitive paths.

> 2. the default behavior would be to use system settings

I'm not sure if `memfs` should "magically" change the way it handles paths; I would prefer it is set explicitly. But happy to hear opinions.

G-Rath:
I might be missing something, but:

> As you probably know, there are 2 options for case sensitivity:

The OSs act the opposite way around from your list: Windows is case insensitive, while UNIX is case sensitive.

![image](https://user-images.githubusercontent.com/3151613/81512725-bf1d1680-9376-11ea-88bd-cbd1dbb2455c.png)

> I'm not sure if memfs should "magically" change the way it handles paths

For now I agree with that - I think `memfs` should be careful adopting "magical" features since they tend to be very "gain this, lose that", but an explicit option could work.

I think in this case actually the OS side should be ignored, since the reason why `fs` is/isnt case-sensitive is because of that; aka it's not Node going "oh you're on Linux", it's instead a result of the underlying API calls V8 makes to the native file system.

Because of that, I think you could make a case for saying something like having the current behavior being the "default", and then implementing a `ignoreCase` option that causes memfs to just call `toLowerCase` on everything?

G-Rath:
On an aside: I think `memfs` should probably favor defaulting to unix behaviours since it already uses the unix file path format (i.e `/my/path/` instead of `C:\my\path`)

piotr-oles:
> I might be missing something, but:
>
> > As you probably know, there are 2 options for case sensitivity:
>
> The OSs act the opposite way around from your list: Windows is case insensitive, while UNIX is case sensitive.

[...]

PR #632, opened 2021-02-09
feat(volume): add caseSensitive option

ajafff:
Defaults to 'true'. If 'false', handle file names case insensitive.
Preserve original file names.

Fixes: #533

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -1054,6 +1054,31 @@ describe('volume', () => {
         expect(typeof vol.promises).toBe('object');
       });
     });
+    describe('caseSensitive: false', () => {
+      it('is case insensitive', () => {
+        const vol = new Volume(undefined, false);
+        vol.mkdirSync('/dir');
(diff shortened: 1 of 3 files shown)


G-Rath:
I'll review this properly when I have a chance, but for now: I think I'd prefer this option to be `ignoreCase` with a default of `false` rather than have an option with a default of `true` as that tends to be easier to reason mentally.

ajafff:
@G-Rath that makes sense. I'll update the PR accordingly.

I noticed that the sorting of `readdir(Sync)` should also respect case sensitivity. I'm currently working on that.

I should probably also update the docs.

tstordyallison:
+1 for this. I'm playing around with using memfs in conjunction with some libraries that are assuming case-insensitivity on Windows platforms, which isn't the case when I'm using memfs! This would be an ideal fix.

jburrow:
+1 Just hit the same issue with memfs on windows

G-Rath:
Closing due to age - feel free to rebase this PR or open a new one if someone wants to pick it back up

PR #681, opened 2021-05-14
fix: prevent callback from triggering twice when callback throws

### LINKED ISSUE #542: Callback was already called.
Hi, I'm using memfs with webpack and its watcher. After rebuild I want to read a file that was just built, but get this error:

```Uncaught exception:  Error: Callback was already called.
    at throwError (C:\...\node_modules\neo-async\async.js:16:11)
    at C:\...\node_modules\neo-async\async.js:2818:7
    at C:\...\node_modules\neo-async\async.js:2830:7
    at done (C:\...\node_modules\neo-async\async.js:3517:9)
    at C:\...\node_modules\webpack\lib\Compiler.js:548:25
    at Immediate.<anonymous> (C:\...\node_modules\memfs\lib\volume.js:684:17)
    at processImmediate (internal/timers.js:456:21)```

### PR DESCRIPTION
**Pull request by privatenumber**
closes #542

### DIFF
--- a/src/__tests__/volume/callback-error.test.ts
+++ b/src/__tests__/volume/callback-error.test.ts
@@ -1,0 +1,15 @@
+jest.useFakeTimers('modern');
+
+// Fixes https://github.com/streamich/memfs/issues/542
+it('should throw error instead of callback', () => {
+  const { Volume } = require('../../volume');
+  const vol = new Volume();
+
+  vol.writeFile('/asdf.txt', 'asdf', 'utf8', (err) => {
+    if (!err) {
+      throw new Error('try to trigger catch');
+    }
+  });
+
+  expect(() => jest.runAllTimers()).toThrow('try to trigger catch');
+});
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/__tests__/volume/callback-error.test.ts [hunk 1]
```diff
@@ -0,0 +1,15 @@
...
+  const vol = new Volume();
+
+  vol.writeFile('/asdf.txt', 'asdf', 'utf8', (err) => {
+    if (!err) {
+      throw new Error('try to trigger catch');
+    }
+  });
+
+  expect(() => jest.runAllTimers()).toThrow('try to trigger catch');
+});
```
privatenumber, line 15: Since `wrapAsync` uses `setImmediate` to invoke the callback that throws an error, the error can't be caught.

The only way to catch it seemed to be by stubbing `setImmediate`, which the `jest.useFakerTimers()` at the top does. This queues all callbacks until `jest.runAllTimers()` is invoked, which runs the queued callbacks synchronously to catch the error being thrown.

`jest.useFakerTimers()`  seemed to stub `setImmediate` for all tests in the file, so I had to put it in a separate file to only influence this specific test.

polkovnikov-ph:
@streamich Could you please review this?

streamich (approved):
Thank you!

PR #732, opened 2021-08-27
fix: global and timers this arg in browser

Mesteery:
This fixes for example `Uncaught TypeError: 'setTimeout' called on an object that does not implement interface Window.` on Firefox.


streamich (approved):
Thank you!

PR #755, opened 2021-09-19
feat: 🎸 add promises.watch() implementation

streamich:


G-Rath (commented):
nice! I think renovate will complain about `callback-to-async-iterator` not being pinned, but have been thinking if it's ok with you I might change the renovate setup to do a lockfile refresh every week to cut down on the noise and number of PRs.

PR #765, opened 2021-10-02
chore: 🤖 migrate tslint setup to eslint.

corwin-of-amber:
It seems that `tslint` is deprecated.

https://github.com/palantir/tslint/issues/4534

I have done a "best effort" to create a compatible `eslint` configuration, utilizing `typescript-eslint`, which is supposed to be `tslint`'s direct successor.

There are more disabled rules now, because they were emitting errors on the current code base. Some you may wish to resolve in the code base, then re-enable.

Cheers 👋


corwin-of-amber:
Should I rebase this?

corwin-of-amber:
I see work on memfs has renewed lately. I am 100% willing to work on this fix, as well as [this one](https://github.com/corwin-of-amber/memfs/commit/8c4fa0f4f0a7583a3e84e6124d42ee396cf4b088) that allows using `buffer` version 6 (which I'm using so I need this). Just signal if you're available.

G-Rath:
@corwin-of-amber sorry for the late response - I'm interested in landing this, if you wouldn't mind rebasing :)

eslint.json
```diff
@@ -0,0 +1,26 @@
+{
```
G-Rath, line 1: Let's stick this in `.eslintrc.js` as that's more conventional

G-Rath (changes requested):

Issue #397, opened 2019-06-16
rename() should work for folders


KutnerBitsrc:
When running .rename() on a folder,
it is expected to rename the folder, and update all file paths in that folder.

But, actually there is no effect. :(

```js
const {vol} = require('memfs')
vol.mkdirSync('/a');
vol.writeFileSync('/a/a.txt', 'hello');
console.log(vol.toJSON()) // {a/a.txt: "hello"}
vol.renameSync('/a/', '/b/')
console.log(vol.toJSON()) // {/a/a.txt: "hello"}
```

corwin-of-amber:
It seems that it changes the directory entries but not the name of the link itself.
```js
> m = require('memfs')
> m.mkdirpSync("/a/b")
> m.writeFileSync("/a/b/c", "d")
> m.renameSync("/a/b", "/a/e")
> m.readdirSync("/a")
[ 'e' ]
> m.readFileSync("/a/e/c")
Uncaught Error: ENOENT: no such file or directory, open '/a/b/c'
    at createError (.../volume.js:131:17)
    at Volume.openLink (.../volume.js:776:19)
...
```

Although directory `/a` contains `e`, the old directory name `b` is still there somehow, and it is picked up by `openLink`.

corwin-of-amber:
Is it possible to make `Link.steps` a property, so that it updates if the parent changes?
https://github.com/streamich/memfs/blob/bb10785965ccc3b7b065e09d70ed8437f4d9dddd/src/node.ts#L257

Otherwise, `steps` has to be recursively updated for children of the moved directory; it is messier, but performance-wise may still be cheaper than recomputing the path on every access.

jorenbroekema:
Hey, is there a workaround for this bug? Currently running into this issue, renaming a directory and following by readFileSync on one of the files in the renamed directory will try to read it from the old directory, and then complain it doesn't exist (which is correct).

I'm glancing over the code seeing if I can help somehow but I'll probably need some directions in order to contribute a fix, as I'm not sure that I understand the nature of the problem.

ronakj:
Just encountered this issue with my tests. Thanks for the fix @jorenbroekema 💯.

jorenbroekema:
You're welcome :)!

PR #795, opened 2021-12-13
fix: recursively sync children steps to fix rename

jorenbroekema:
fixes https://github.com/streamich/memfs/issues/397

I'm not familiar with the repo, just diving into it just now so forgive me if this approach is not what you guys had in mind. I didn't really see a trivial way around it other than creating a setter and recursively syncing children to parent.

When "steps" prop is set to a new value (e.g. in `renameBase`), it also goes through our children and sets their steps to synchronize with their parent which has changed.

### DIFF
--- a/src/node.ts
+++ b/src/node.ts
@@ -239,7 +239,7 @@ export class Link extends EventEmitter {
   children: { [child: string]: Link | undefined } = {};
 
   // Path to this node as Array: ['usr', 'bin', 'node'].
-  steps: string[] = [];
+  private _steps: string[] = [];
 
   // "i-node" of this hard link.
   node: Node;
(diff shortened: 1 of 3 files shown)


corwin-of-amber:
Looks like a great fix!

src/node.ts
```diff
@@ -250,11 +250,26 @@ export class Link extends EventEmitter {
...
+  name: string;
+
+  get steps() {
+    return this._steps;
+  }
+
+  // Recursively sync children steps, e.g. in case of dir rename
+  set steps(val) {
+    this._steps = val;
+    Object.entries(this.children).forEach(([, child]) => {
```
G-Rath, line 262:
```suggestion
    Object.values(this.children).forEach((child) => {
```

nit: I find `for...of` tends to be nicer than `forEach`, but that's not a blocker.

src/node.ts
```diff
@@ -239,7 +239,7 @@ export class Link extends EventEmitter {
   children: { [child: string]: Link | undefined } = {};
 
   // Path to this node as Array: ['usr', 'bin', 'node'].
-  steps: string[] = [];
+  _steps: string[] = [];
```
G-Rath, line 242:
```suggestion
  private _steps: string[] = [];
```

G-Rath (commented):
I think this looks good, but will have a better look later this week when I've got more bandwidth :)

src/node.ts
jorenbroekema: Adjusted

src/node.ts
jorenbroekema: Adjusted

streamich (approved):
Thank you!

PR #838, opened 2022-05-17
fix: ensure js file is copied to expected place

### LINKED ISSUE #837: Cannot find module './getBigInt'
Apologies if this has been answered before or I'm just not understanding it, but `memfs@^3.4.1` as a peer dependency of `webpack-dev-middleware` has an issue where the lib/Stats.js fails to import `./getBigInt` – looking through `node_modules/memfs` the getBigInt.js file is located at `lib/src/getBigInt.js` instead of `lib/getBigInt.js`

Again, sorry if this has been addressed before or I'm not doing it right. Just wanted to see if this has been patched somewhere and I'm a little confused. For now I've reverted to 3.4.1 in my project's `package-lock.json` so we're alright for now, but wanted to post this issue in here in case it was at all helpful. Thanks!

G-Rath:
`cpy-cli` was upgraded to a new major version by #818 which has since resulted in `getBigInt.js` being copied into `lib/src/getBigInt.js` instead of `lib/getBigInt.js` which has broken the latest release (#837).

I can't see what the breaking change in `cpy-cli` that is causing this, but frankly we've only got `getBigInt.js` to support lower versions of node so we shouldn't be expecting to copy any other js file & so I've just replaced it with `cp` which is available on standard Linux & MacOS systems (and I believe PowerShell on Windows).

This'll unbreak the library for the meantime at least - we can always add `cpy-cli` back in future.

Resolves #837

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   ],
   "scripts": {
     "clean": "rimraf lib types",
-    "build": "tsc -p . && cpy src/*.js lib",
+    "build": "tsc -p . && cp src/getBigInt.js lib/",
     "test": "jest --maxWorkers 2",
     "test:coverage": "jest --coverage",
     "test:watch": "jest --watch",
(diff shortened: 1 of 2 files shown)


conlanpatrek:
Confirmed, thank you for clearing this so quickly.

PR #847, opened 2022-06-07
fix: dont patch `getuid` and `getgid` on process anymore #809

### LINKED ISSUE #809: Side effect when mocking `process.getuid()` under Windows
Recently I have found that when I imported `memfs/lib/process.js` and used `gulp` to copy a directory with a subdirectory, I got an error `Error: EPERM: operation not permitted, futime`.

```
dest
test
│   index.js
└───sub
    │   index.js
```

Here is the snippet to show how to reproduce the problem:

```js
// test/index.js
const path = require('path');
const gulp = require('gulp');

// require('memfs/lib/process.js');
process.getuid = process.getuid || (() => 0);
gulp.src(['**/*']).pipe(gulp.dest(path.resolve(__dirname, '../dest')))
```

It seems that it is a mocked `uid` resulting in side effects when trying to `futime`.

The temporary workaround is to revert it within the current process:

```js
require('memfs');
process.platform === 'win32' && delete process.getuid;
```

aleen42:
Resolves #809

### DIFF
--- a/src/__tests__/process.test.ts
+++ b/src/__tests__/process.test.ts
@@ -7,8 +7,8 @@ describe('process', () => {
       expect(typeof _process).toBe('object');
     });
     it('.getuid() and .getgid()', () => {
-      expect(typeof proc.getuid()).toBe('number');
-      expect(typeof proc.getgid()).toBe('number');
+      expect(typeof proc.getuid?.() ?? 0).toBe('number');
+      expect(typeof proc.getgid?.() ?? 0).toBe('number');
     });
     it('.cwd()', () => {
       expect(typeof proc.cwd()).toBe('string');
--- a/src/node.ts
+++ b/src/node.ts
@@ -6,6 +6,8 @@ import { EventEmitter } from 'events';
 import Stats from './Stats';
 
 const { S_IFMT, S_IFDIR, S_IFREG, S_IFLNK, O_APPEND } = constants;
+const getuid = (): number => process.getuid?.() ?? 0;
+const getgid = (): number => process.getgid?.() ?? 0;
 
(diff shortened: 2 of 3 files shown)


src/node.ts
```diff
@@ -17,7 +17,6 @@ export class Node extends EventEmitter {
   ino: number;
 
   // User ID and group ID.
-  uid: number = process.getuid();
```
G-Rath, line 20: This shouldn't be removed as it's public + used by the rest of this class.

I think let's abstract the falling back behaviour into a small helper function within this file and then we can call that everywhere, e.g.

```
const getuid = (): number => process.getuid?.() ?? 0

export class Node extends EventEmitter {
  // ...
  uid: number = getuid();
}
```

G-Rath (changes requested):
Looks good, just need to restore the `uid` property.

Would you be open to also replacing `getgid` in this PR too, since that change is pretty much the same thing?

src/__tests__/process.test.ts
```diff
@@ -7,8 +7,8 @@ describe('process', () => {
       expect(typeof _process).toBe('object');
     });
     it('.getuid() and .getgid()', () => {
-      expect(typeof proc.getuid()).toBe('number');
-      expect(typeof proc.getgid()).toBe('number');
+      expect(typeof proc.getuid?.() ?? 0).toBe('number');
+      expect(typeof proc.getgid?.() ?? 0).toBe('number');
     });
```
G-Rath, line 12: nit: we should remove this test entirely as it's not really covering anything useful now

G-Rath (approved):

PR #862, opened 2022-10-04
Add ESM support

### LINKED ISSUE #861: Publish ESM with import syntax
In order to use vitest to mock dependencies of memfs, the dependencies have to use native JavaScript imports.

This can be used as a workaround to #327.

steinybot:
Fixes #861

I tried to keep the CJS module the same.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,19 +3,30 @@
   "version": "3.4.7",
   "description": "In-memory file-system with Node's fs API.",
   "main": "lib/index.js",
+  "module": "lib/index.mjs",
   "types": "lib/index.d.ts",
   "files": [
     "lib"
   ],
+  "exports": {
+    ".": {
+      "require": "./lib/index.js",
+      "import": "./lib/index.mjs"
+    },
+    "./lib/promises": {
+      "require": "./lib/promises.js",
+      "import": "./lib/promises.mjs"
+    }
+  },
   "scripts": {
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -3,13 +3,24 @@
   "version": "3.4.7",
   "description": "In-memory file-system with Node's fs API.",
   "main": "lib/index.js",
+  "module": "lib/index.mjs",
   "types": "lib/index.d.ts",
   "files": [
     "lib"
   ],
+  "exports": {
```
steinybot, line 11: Should everything go in here?

G-Rath (changes requested):
I don't feel comfortable with this change because it looks to be very breaking and this family of modules is somewhat fragile (i.e we can't even publish new versions of `unionfs` right now), so I think this'll have to be a no from me at this point.

I can try and look into ESM support at some point, but cannot give a timeline on when that might actually be.

G-Rath:
Those are all pretty serious changes and I've not got the time to actively work on this library right now to ensure everything goes smoothly; we'd also need to make sure other libraries like `unionfs` are compatible, which again we don't have the ability to publish right now.

Sorry but I don't think this will happen anytime soon. I can look into supporting overriding `path` via DI which sounds like it would unblock the Windows file path issue

G-Rath:
Closing per my previous comments

PR #866, opened 2022-10-26
fix: support calling `utimes` on a directory

### LINKED ISSUE #391: "utimes" doesn't work for directories: "EISDIR: illegal operation on a directory, open ..." error
`utimes` calls `utimesBase` which uses `openSync` call internally but `openSync` call on a directory obviously throws `EISDIR` error.

https://github.com/streamich/memfs/blob/674a1e78469e1e9e766db7e3160538e958fdd97b/src/volume.ts#L1771-L1778

williamstein:
- "EISDIR: illegal operation on a directory, open ..." error
- https://github.com/streamich/memfs/issues/391

Resolves #391

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -961,10 +961,17 @@ describe('volume', () => {
     });
     describe('.utimesSync(path, atime, mtime)', () => {
       const vol = new Volume();
+      vol.mkdirSync('/foo');
       it('Set times on file', () => {
-        vol.writeFileSync('/lol', '12345');
-        vol.utimesSync('/lol', 1234, 12345);
-        const stats = vol.statSync('/lol');
+        vol.writeFileSync('/foo/lol', '12345');
+        vol.utimesSync('/foo/lol', 1234, 12345);
+        const stats = vol.statSync('/foo/lol');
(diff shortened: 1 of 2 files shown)


src/__tests__/volume.test.ts
```diff
@@ -968,10 +968,17 @@ describe('volume', () => {
...
-        vol.writeFileSync('/lol', '12345');
-        vol.utimesSync('/lol', 1234, 12345);
-        const stats = vol.statSync('/lol');
+        vol.writeFileSync('/foo/lol', '12345');
+        vol.utimesSync('/foo/lol', 1234, 12345);
+        const stats = vol.statSync('/foo/lol');
+        expect(Math.round(stats.atime.getTime() / 1000)).toBe(1234);
+        expect(Math.round(stats.mtime.getTime() / 1000)).toBe(12345);
+      });
+      it('Sets times on a directory (see https://github.com/streamich/memfs/issues/391)', () => {
```
G-Rath, line 979: I think this test case is obvious enough that we don't need to tie it to a specific issue:

```suggestion
      it('Sets times on a directory', () => {
```

G-Rath (approved):

PR #867, opened 2022-10-26
fix: return `/` when calling `realpathSync` with `/`

### LINKED ISSUE #863: realpathSync('/') returns invalid result (edge case bug)
This is easy to reproduce:

```js
> memfs = require('memfs');
> memfs.vol.fromJSON({'./a':'a'}, '/');
> memfs.vol.realpathSync('/')
''
```
Notice that the output is an empty string, which isn't a valid path.  This bug is caused by getPath() in node.ts:
```ts
  getPath(): string {
    return this.steps.join(SEP);
  }
```
The problem is the edge case of join when `this.steps` is `['']`, i.e., a list with a single empty string.

The following is what I would expect as output in this case:
```js
> require('fs').realpathSync('/')
'/'
```

Perhaps a good fix would be:

```ts
  getPath(): string {
    const path = this.steps.join(SEP);
    return path ? path : "/";
  }
```

williamstein:
-  https://github.com/streamich/memfs/issues/863

Resolves #863

### DIFF
--- a/src/__tests__/volume/realpathSync.test.ts
+++ b/src/__tests__/volume/realpathSync.test.ts
@@ -1,6 +1,6 @@
 import { create } from '../util';
 
-describe('.realpath(...)', () => {
+describe('.realpathSync(...)', () => {
   it('works with symlinks, #463', () => {
     const vol = create({});
     vol.mkdirSync('/a');
@@ -11,4 +11,9 @@ describe('.realpath(...)', () => {
     const path = vol.realpathSync('/a/b/index.js');
     expect(path).toBe('/c/index.js');
   });
+});
+
+describe("edge case -- realpathSync('/') returns '/'", () => {
+  const vol = create({'./a':'a'});
+  expect(vol.realpathSync('/')).toBe('/');
 });
--- a/src/__tests__/volume/realpathSync.test.ts
+++ b/src/__tests__/volume/realpathSync.test.ts
@@ -14,6 +14,6 @@ describe('.realpathSync(...)', () => {
 });
 
 describe("edge case -- realpathSync('/') returns '/'", () => {
-  const vol = create({'./a':'a'});
+  const vol = create({ './a': 'a' });
   expect(vol.realpathSync('/')).toBe('/');
 });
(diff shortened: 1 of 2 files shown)


src/__tests__/volume/realpathSync.test.ts
```diff
@@ -12,3 +12,8 @@ describe('.realpath(...)', () => {
     expect(path).toBe('/c/index.js');
   });
 });
+
+describe("edge case -- realpathSync('/') returns '/'", () => {
```
G-Rath, line 16: Similar to [your other PR](https://github.com/streamich/memfs/pull/870#discussion_r1008749448), `describe` isn't a test case - since we've already got a `describe` for `realpathSync` (though I'm not quite sure why we've got a dedicated file for it, but oh well), let's just stick the test in that with a more descriptive title:

```
describe('.realpathSync(...)', () => {
  it('works with symlinks, #463', () => {
    // ...
  });

  it('returns the root correctly', () => {
    // ...
  });
});
```

src/volume.ts
```diff
@@ -1493,7 +1493,8 @@ export class Volume {
     const realLink = this.getResolvedLink(steps);
     if (!realLink) throw createError(ENOENT, 'realpath', filename);
 
-    return strToEncoding(realLink.getPath(), encoding);
+    const path = realLink.getPath();
+    return strToEncoding(path ? path : '/', encoding);
```
G-Rath, line 1497: we can just use `||` here right?

```suggestion
    return strToEncoding(realLink.getPath() || '/', encoding);
```

G-Rath (changes requested):

src/volume.ts
G-Rath: btw fwiw this behavior technically is not correct for Windows, but I think this is still better than the current behaviour so happy to land it

image

williamstein:
OK, I fixed these tests.

G-Rath:
Awesome! Have you seen [this comment](https://github.com/streamich/memfs/pull/867#discussion_r1008750149)? otherwise this looks good to go

williamstein:
I saw the comment, but I'm also not sure what to do about it, or what all of the other MS Windows conventions are with memfs, so I don't feel qualified to address it.  I may come back to it later, since Windows support is a longterm goal of my project (https://cowasm.org).

G-Rath:
I was actually referring to the original comment/suggestion about using `||` to avoid the local variable :)

For the windows comment, thats completely fine though in future I'd recommend putting a reaction such as a thumbs up to indicate you've seen the comment and acknowledge it without any further comment yourself

G-Rath (approved):

PR #870, opened 2022-10-26
fix: support calling `chmod` on a directory

### LINKED ISSUE #558: `chmodSync` throws EISDIR for directories
Running `chmodSync` on a directory fails with EISDIR:

```js
var { fs } = require("memfs")

fs.mkdirSync('/foo')
fs.chmodSync('/foo', 0o666)
```

The error is: `Error: EISDIR: illegal operation on a directory, open '/foo'`.

(Try it: https://runkit.com/dabbott/memfs-chmodsync/1.0.0)

The same calls seem to work with node's built-in `fs`, as long as it's a directory where you have permission to write, so I'm guessing this is an issue with `chmodSync`?

If this is a bug I can try to submit a PR. I haven't looked through the code yet, but seems like it'd be a quick change.

---

Running `chmod` (not sync) works. Also, running `openSync` and then `fchmodSync` works, as a workaround:

```js
var { fs } = require("memfs")

fs.mkdirSync('/foo')
fs.fchmodSync(fs.openSync('/foo', 'r'), 0o666)
```

williamstein:
- See https://github.com/streamich/memfs/issues/558

The fix for this is similar to the others I just submitting, so here it is.

Resolves #558

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -1103,6 +1103,14 @@ describe('volume', () => {
         expect(typeof vol.promises).toBe('object');
       });
     });
+    describe('.chmodSync(directory) works (see https://github.com/streamich/memfs/issues/558', () => {
+      const vol = new Volume();
+      vol.mkdirSync('/dir');
+      vol.chmodSync('/dir', 0o666);
+      expect(vol.statSync('/dir').mode.toString(8)).toBe('40666');
+      vol.chmodSync('/dir', 0o777);
+      expect(vol.statSync('/dir').mode.toString(8)).toBe('40777');
+    });
   });
   describe('StatWatcher', () => {
     it('.vol points to current volume', () => {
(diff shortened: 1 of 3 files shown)


src/__tests__/volume.test.ts
```diff
@@ -1110,6 +1110,14 @@ describe('volume', () => {
         expect(typeof vol.promises).toBe('object');
       });
     });
+    describe('.chmodSync(directory) works (see https://github.com/streamich/memfs/issues/558', () => {
```
G-Rath, line 1113: `describe` blocks are not proper test cases, so this isn't running as you probably expect - let's lay it out like this:

```
describe('.chmodSync(path, mode)', () => {
  it('works with directories', () => {
    // ...
  });

  it('works with files', () => {
    // ...
  });
});
```

and lets put it above the `.promises`, just so its with the rest of the function tests

G-Rath (changes requested):

williamstein:
I've restructured the tests.

G-Rath (approved):

PR #875, opened 2022-10-28
fix: truncate to larger size

toyobayashi:


G-Rath:
Could you give a bit more a description/story about this change? why is it a problem, how did you discover it, what does Node do? etc

toyobayashi:
@G-Rath 

```js
const fs = require('fs')

const fd = fs.openSync('./test', 'w');
fs.writeFileSync(fd, '12345');
console.log(fs.readFileSync('./test', 'utf8') === '12345');
fs.truncateSync('./test', 10);
console.log(fs.readFileSync('./test', 'utf8') === '12345\0\0\0\0\0');
```

According to https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsftruncatefd-len-callback

> If the file previously was shorter than len bytes, it is extended, and the extended part is filled with null bytes ('\0'):

toyobayashi:
I'm using memfs in browser, before this change, `ftruncateSync` just throws RangeError if truncate to a larger size

image

G-Rath (approved):

PR #880, opened 2022-11-12
feat(volume): now handling `Buffer` as volume content

shorwood:
## Context

Hi ! While trying to inject some `Buffer` into a `memfs` volume from the `fromJSON` method, I noticed it creates a virtual folder. I expected the method to accept `Buffer` as content but I noticed only `string` are "allowed".

This PR resolves cases when you want to inject raw binary data into your volumes. For example when you have some test archives to validate.

## Actions

I have written a small patch to fix the issue. I tried to make as few changes as possible. It comes with a new test and some adjustments to the documentation.

I hope it's enough for a quick merge. If not, feel free to share adjustments you would like to be done. I'll try to be as reactive as possible.

Best regards,

Stanley.

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -266,6 +266,16 @@ describe('volume', () => {
         expect(stat.isDirectory()).toBe(true);
         expect(vol.readdirSync('/dir')).toEqual([]);
       });
+
+      it('supports using buffers for file content', () => {
+        const vol = new Volume();
+        const text = 'bip-boup';
+        const buffer = Buffer.from(text, 'utf-8');
+        vol.fromJSON({ '/buffer': buffer });
+        expect(vol.toJSON()).toStrictEqual({ '/buffer': text });
+        expect(vol.readFileSync('/buffer')).toStrictEqual(buffer);
+        expect(vol.readFileSync('/buffer', 'utf-8')).toStrictEqual(text);
+      });
     });
 
     describe('.fromNestedJSON(nestedJSON[, cwd])', () => {
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -186,13 +186,13 @@ function validateGid(gid: number) {
 }
 
 // ---------------------------------------- Volume
-type DirectoryContent = string | null;
+type DirectoryContent = string | Buffer | null;
 
-export interface DirectoryJSON {
-  [key: string]: DirectoryContent;
+export interface DirectoryJSON<T extends DirectoryContent = DirectoryContent> {
+  [key: string]: T;
 }
-export interface NestedDirectoryJSON {
-  [key: string]: DirectoryContent | NestedDirectoryJSON;
+export interface NestedDirectoryJSON<T extends DirectoryContent = DirectoryContent> {
+  [key: string]: T | NestedDirectoryJSON;
 }
 
 function flattenJSON(nestedJSON: NestedDirectoryJSON): DirectoryJSON {
@@ -204,7 +204,7 @@ function flattenJSON(nestedJSON: NestedDirectoryJSON): DirectoryJSON {
 
       const joinedPath = join(pathPrefix, path);
 
-      if (typeof contentOrNode === 'string') {
+      if (typeof contentOrNode === 'string' || contentOrNode instanceof Buffer) {
         flatJSON[joinedPath] = contentOrNode;
       } else if (typeof contentOrNode === 'object' && contentOrNode !== null && Object.keys(contentOrNode).length > 0) {
         // empty directories need an explicit entry and therefore get handled in `else`, non-empty ones are implicitly considered
@@ -524,7 +524,7 @@ export class Volume implements FsCallbackApi {
     });
   }
 
-  private _toJSON(link = this.root, json = {}, path?: string): DirectoryJSON {
+  private _toJSON(link = this.root, json = {}, path?: string): DirectoryJSON<string | null> {
     let isEmpty = true;
 
     let children = link.children;
@@ -566,7 +566,7 @@ export class Volume implements FsCallbackApi {
     return json;
   }
 
-  toJSON(paths?: PathLike | PathLike[], json = {}, isRelative = false): DirectoryJSON {
+  toJSON(paths?: PathLike | PathLike[], json = {}, isRelative = false): DirectoryJSON<string | null> {
     const links: Link[] = [];
 
     if (paths) {
@@ -593,7 +593,7 @@ export class Volume implements FsCallbackApi {
 
       filename = resolve(filename, cwd);
 
-      if (typeof data === 'string') {
+      if (typeof data === 'string' || data instanceof Buffer) {
         const dir = dirname(filename);
         this.mkdirpBase(dir, MODE.DIR);
 
--- a/docs/node/reference.md
+++ b/docs/node/reference.md
@@ -71,6 +71,7 @@ vol.fromJSON(
   {
     './index.js': '...',
     './package.json': '...',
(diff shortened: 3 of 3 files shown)


src/__tests__/volume.test.ts
```diff
@@ -263,6 +263,17 @@ describe('volume', () => {
         expect(stat.isDirectory()).toBe(true);
         expect(vol.readdirSync('/dir')).toEqual([]);
       });
+
+      it('Files as buffer', () => {
```
G-Rath, line 267: Can you add an `fs.readFile` call to ensure the return from that is valid too?

src/__tests__/volume.test.ts
```diff
@@ -263,6 +263,17 @@ describe('volume', () => {
...
+        const vol = new Volume();
+        const json = {
+          '/buffer': Buffer.from('bip-boup', 'utf-8'),
+        };
+        vol.fromJSON(json);
+        expect(vol.toJSON()).toEqual({
```
G-Rath, line 273:
```suggestion
        expect(vol.toJSON()).toStrictEqual({
```

(I know we're not using it in other tests, but we should use it going forward)

src/volume.ts
```diff
@@ -516,7 +516,7 @@ function validateGid(gid: number) {
 }
 
 // ---------------------------------------- Volume
-type DirectoryContent = string | null;
+type DirectoryContent = string | Buffer | null;
```
G-Rath, line 519: hmm something feels a bit off to me about this change, because it represents both the input and output of `fromJson` and `toJson` respectively yet in your test it looks like giving a Buffer to `fromJson` results in a string when calling `toJson`.

Can you check why that is and if it's the correct behaviour then we'll need to modify the return type of `toJson` to reflect that

G-Rath (changes requested):

src/volume.ts
shorwood: I agree, it was off. `toJSON` only returns a map of utf-8 strings.

Here are the changes i've made to resolve the issue:
- Transformed `DirectoryJSON` and `DirectoryNestedJSON` as optionnal generics to allow for type narrowing when needed.
- Changed the return type of `toJSON` and `_toJSON` to reflect the correct behavior.

src/__tests__/volume.test.ts
shorwood: Done :)

src/__tests__/volume.test.ts
shorwood: Done.

src/volume.ts
G-Rath: > Changed the return type of vol.toJSON and vol._toJSON to reflect the correct behavior:

I don't think that is the correct behaviour for the very reason you want to enable using `Buffer` for `fromJSON` in the first place: binary data cannot be represented as a string - i.e. this will produce an invalid zip:

```
const fs = require('fs');

// assume this exists on disk and is a plain zip with one file
const myZip = fs.readFileSync('input.zip');

const vol = new Volume();

vol.fromJSON({ 'my.zip': myZip });

fs.writeFileSync('output.zip', vol.toJSON()['my.zip']);
```

shorwood: You're right, the changes ignores this case.

I skipped over it because from my point of view (and I might be wrong), `vol.toJSON` is used as debug debug utility, not as an output method. I justify this because the snippet bellow already produces the same invalid zip :

```diff
const { Volume } = require('memfs');
const fs = require('fs');

// assume this exists on disk and is a plain zip with one file
const myZip = fs.readFileSync('input.zip');

const vol = new Volume();

-- vol.fromJSON({ 'my.zip': myZip });
++ vol.writeFileSync('my.zip', myZip);

fs.writeFileSync('output.zip', vol.toJSON()['my.zip']);
```

To me, the zip would be valid if we did:

```diff
const { Volume } = require('memfs');
const fs = require('fs');

// assume this exists on disk and is a plain zip with one file
const myZip = fs.readFileSync('input.zip');

const vol = new Volume();

vol.fromJSON({ 'my.zip': myZip });

-- fs.writeFileSync('output.zip', vol.toJSON()['my.zip']);
++ fs.writeFileSync('output.zip', vol.readFileSync('/my.zip'));
```

## Food for thoughts
But as you pointed out, now that input and output types are not symetrical, this might bring confusion.

One way of reducing this confusion would be to add an optionnal `encoding?: BufferEncoding` parameter to the `vol.toJSON`. Therefore allowing us to explicitly specify the return type of `vol.toJSON`

```ts
vol.toJSON() // { 'file.txt': [Buffer] }
vol.toJSON('utf-8') // { 'file.txt': 'Hello World !' }
```

However, this would bring a breaking change that might have a high impact on existing dependent codebases. One solution to that would be to default `encoding` to `utf-8`. But is the effort worth it and are users going to be using this parameter that ofter ?

## What are your thoughts ?
- Should we assume `vol.toJSON` is just a debug utility and ignore the use case.
- Or should we allow the user to explicitly set the output type of `vol.toJSON`.

shorwood: To clarify and avoid any misunderstandings:

The return type of `toJSON` is still the same as before: `{ [x: string]: string | null }`

G-Rath: Good point about `writeFile` - I think the right solution would be something along the lines of that `encoding` parameter (except I'd probably make it a boolean like `asBuffers` since I don't think there's any other options that would make sense); and that means we can land this as-is 🎉

shorwood: Havent thought of that ! `asBuffer` would be optionnal, fit the need and would not be a breaking change. I'll make the required changes; However, I'm a bit hesitant on where to add this parameter.

Maybe a forth parameter ?

```ts
toJSON(paths?: PathLike | PathLike[], json = {}, isRelative = false, asBuffer = false): DirectoryJSON<string | null>
```

G-Rath: Cool but lets do it in a new PR because it's easily a separate feature and deserves its own line in the changelog :)

(and yes for now I'd just add a fourth parameter - an `options` object would be better, but overloading is a bit of a hassle and we've not gotten a lot of requests for this so we can improve it later)

shorwood: Here it is: #881

G-Rath (changes requested):
You need to apply prettier

src/__tests__/volume.test.ts
```diff
@@ -263,6 +263,16 @@ describe('volume', () => {
         expect(stat.isDirectory()).toBe(true);
         expect(vol.readdirSync('/dir')).toEqual([]);
       });
+
+      it('Files as buffer', () => {
```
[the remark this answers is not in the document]
shorwood: Resolved: I applied suggested changes.

shorwood:
> You need to apply prettier

Done.

G-Rath:
@shorwood are you able to rebase this please?

shorwood:
> @shorwood are you able to rebase this please?

Hi, it's done. Sorry for the close/reopen, I had to sync and cherry pick because rebase was taking too long on my PC. I have also taken the opportunity to merge the commit from #881 (since they are very similar) and reword them a bit for better understanding.

Awaiting your feedback.

shorwood:
> @shorwood are you able to rebase this please?

Hi, it's done. Sorry for the close/reopen, I had to sync and cherry pick because rebase was taking too long on my PC. I have also taken the opportunity to merge the commit from #881 (since they are very similar) and reword them a bit for better understanding.

Awaiting your feedback.

G-Rath (approved):

G-Rath (approved):

PR #882, opened 2022-11-14
fix: `mkdir` return value

### LINKED ISSUE #546: Typing mismatch between fs.mkdir and memfs.mkdir
Firstly and most importantly, huge thanks for this package, it's awesome.

I tried to drop in the output from `createFsFromVolume()` instead of the native `fs` and I hit a snag. There's a minor typing mismatch.

```
Conversion of type 'IFs' to type 'FS' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  The types returned by 'promises.mkdir(...)' are incompatible between these types.
    Type 'Promise<void>' is not comparable to type 'Promise<string>'.
      Type 'void' is not comparable to type 'string'.ts(2352)
```

Looking into this a little, the node [fs.mkdir](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback) and [fs.promises.mkdir](https://nodejs.org/api/fs.html#fs_fspromises_mkdir_path_options) optionally return the path to the first directory created when the `recursive: true` option is set. While `memfs.mkdir` always returns `void`.

toyobayashi:
https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsmkdirpath-options-callback

> The callback is given a possible exception and, if recursive is true, the first directory path created, (err[, path]). path can still be undefined when recursive is true, if no directory was created.

https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsmkdirsyncpath-options

> Synchronously creates a directory. Returns undefined, or if recursive is true, the first directory path created.

Resolves #546

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -433,7 +433,7 @@ if (isWin) {
 
 export function filenameToSteps(filename: string, base?: string): string[] {
   const fullPath = resolve(filename, base);
-  const fullPathSansSlash = fullPath.substr(1);
+  const fullPathSansSlash = fullPath.substring(1);
   if (!fullPathSansSlash) return [];
   return fullPathSansSlash.split(sep);
 }
@@ -728,7 +728,7 @@ export class Volume {
 
   // Generates 6 character long random string, used by `mkdtemp`.
   genRndStr() {
-    const str = (Math.random() + 1).toString(36).substr(2, 6);
+    const str = (Math.random() + 1).toString(36).substring(2, 8);
     if (str.length === 6) return str;
     else return this.genRndStr();
   }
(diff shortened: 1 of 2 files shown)


toyobayashi:
not sure why Jest runs timeout on macOS CI

toyobayashi:
Perhaps this PR can fix #546

src/volume.ts
```diff
@@ -1920,8 +1922,11 @@ export class Volume {
    * @param modeNum
    */
   private mkdirpBase(filename: string, modeNum: number) {
-    const steps = filenameToSteps(filename);
+    const fullPath = resolve(filename);
+    const fullPathSansSlash = fullPath.substr(1);
```
G-Rath, line 1926: `substr` is deprecated - can we replace this with `substring`?

toyobayashi: Done.

G-Rath (approved):

Issue #889, opened 2022-12-19
Creating a file in a directory watchFile does not trigger listener


vangie:
The following is a memfs test I wrote.
```typescript
    test("memfs watchFile", async () => {
        const fs: FileSystem = createFsFromVolume(new Volume());
        fs.mkdirSync("/test_dir/");
        const watchListener = jest.fn();
        try {
            fs.watchFile("/test_dir/", { persistent: true, interval: 50 }, watchListener);
            fs.writeFileSync("/test_dir/1.txt", "test");
            await delay(60);
            const calls = watchListener.mock.calls;
            expect(calls.length).toBe(1);
            expect(calls[0][0]).toMatchObject({ nlink: 3 });
            expect(calls[0][1]).toMatchObject({ nlink: 2 });
        } finally {
            fs.unwatchFile("/test_dir/");
            // fs.rmSync("/test_dir/", { recursive: true, force: true });
        }
    });
```
The result is as follows
```bash
    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 0

      37 |             await delay(60);
      38 |             const calls = watchListener.mock.calls;
    > 39 |             expect(calls.length).toBe(1);
         |                                  ^
      40 |             expect(calls[0][0]).toMatchObject({ nlink: 3 });
      41 |             expect(calls[0][1]).toMatchObject({ nlink: 2 });
      42 |         } finally {

      at Object.<anonymous> (test/file-system.spec.ts:39:34)
```

However, testing with node's fs gives the expected results
```typescript
import fs from "fs";

    test("fs watchFile", async () => {
        fs.mkdirSync("/tmp/watch_test");
        const watchListener = jest.fn();
        try {
            fs.watchFile("/tmp/watch_test", { persistent: true, interval: 50 }, watchListener);
            fs.writeFileSync("/tmp/watch_test/1.txt", "test");
            await delay(60);
            const calls = watchListener.mock.calls;
            expect(calls.length).toBe(1);
            expect(calls[0][0]).toMatchObject({ nlink: 3 });
            expect(calls[0][1]).toMatchObject({ nlink: 2 });
        } finally {
            fs.unwatchFile("/tmp/watch_test");
            fs.rmSync("/tmp/watch_test", { recursive: true, force: true });
        }
    });
```

I found that on my mac m1, the number of nlink in the folder increases with the number of subfiles, while in memfs the number is always 1. Neither the ctime nor the mtime of the parent directory has changed.

```bash
[...]

PR #891, opened 2022-12-21
fix: ensure metadata is updated correctly so that `watchFile` is correctly triggered

### LINKED ISSUE #901: When a file is changed in a memfs volume, its ctime does not update.
mtime and atime both seem to behave as expected, but ctime does not.

Seems like it could be related to #891 and #889

Here's a test I wrote that fails when using a memfs volume but passes when using the real 'fs':

```javascript
const fs = require('fs');
const path = require('path');
const assert = require('assert');

describe('memfs ctime', async () => {

  it('should change ctime', async () => {
    const vol = new Volume();
    
    const _fs_ = vol; // test fails
    // const _fs_ = fs; // test passes
    
    const file = '/tmp/test-memfs-ctime.txt';
    
    // Wait 1ms between actions to be sure the times change.
    const sleep = () => new Promise(resolve => setTimeout(resolve, 1));

    // Create the file if it doesn't exist.
    _fs_.mkdirSync(path.dirname(file), {recursive: true});
    _fs_.closeSync(_fs_.openSync(file, 'a'));
    
    // Get the ctime before any changes.
    let initial = _fs_.statSync(file).ctimeMs;
    
    // Get the ctime after writing to the file.
    await sleep();
    _fs_.writeFileSync(file, 'testing memfs ctime', 'utf8');
    let afterWrite = _fs_.statSync(file).ctimeMs;

    // Get the ctime after changing the file's atime and mtime.
    await sleep();
    let oneSecondAgo = new Date(Date.now() - 1000);
    _fs_.utimesSync(file, oneSecondAgo, oneSecondAgo);
    let afterMtimeChange = _fs_.statSync(file).ctimeMs;
    
    assert(afterWrite > initial, 'ctime should be greater after write than before');
    assert(afterMtimeChange > afterWrite, 'ctime should be greater after utimes change than before');
  });

});
```

vangie:
1. folder's nlink default value is 2
2. add subfolder nlink++
3. atime updated
4. ctime updated after uid/gid/atime/mtime/perm/nlink updated

Resolves https://github.com/streamich/memfs/issues/889
Resolves #901

### DIFF
--- a/src/__tests__/node.test.ts
+++ b/src/__tests__/node.test.ts
@@ -1,4 +1,4 @@
-import { Node } from '../node';
+import { Node, Link } from '../node';
 import { constants } from '../constants';
 
 describe('node.ts', () => {
@@ -68,6 +68,32 @@ describe('node.ts', () => {
       node.chmod(0o600);
       expect(node.perm).toBe(0o600);
       expect(node.isFile()).toBe(true);
+    });
+    describe.each([['uid'], ['gid'], ['atime'], ['mtime'], ['perm'], ['nlink']])('ctime changes', field => {
+      it(`set ${field}`, () => {
+        const node = new Node(1);
+        const oldCtime = node.ctime;
+        node[field] = 1;
+        const newCtime = node.ctime;
+        expect(newCtime !== oldCtime).toBeTruthy();
+      });
+    });
+
+    describe.each([
+      ['getString', []],
+      ['getBuffer', []],
+      ['read', [Buffer.alloc(0)]],
+    ])('atime changes', (method, args) => {
+      it(`${method}()`, () => {
+        const node = new Node(1);
+        const oldAtime = node.atime;
+        const oldCtime = node.ctime;
+        node[method](...args);
+        const newAtime = node.atime;
+        const newCtime = node.ctime;
+        expect(newAtime !== oldAtime).toBeTruthy();
+        expect(newCtime !== oldCtime).toBeTruthy();
+      });
     });
   });
 });
--- a/src/__tests__/node.test.ts
+++ b/src/__tests__/node.test.ts
@@ -60,6 +60,16 @@ describe('node.ts', () => {
         node.read(buf, 0, 1, 1);
         expect(buf.equals(Buffer.from([2]))).toBe(true);
       });
+      it('updates the atime and ctime', () => {
+        const node = new Node(1);
+        const oldAtime = node.atime;
+        const oldCtime = node.ctime;
+        node.read(Buffer.alloc(0));
+        const newAtime = node.atime;
+        const newCtime = node.ctime;
+        expect(newAtime).not.toBe(oldAtime);
+        expect(newCtime).not.toBe(oldCtime);
+      });
     });
     describe('.chmod(perm)', () => {
       const node = new Node(1);
@@ -69,30 +79,38 @@ describe('node.ts', () => {
       expect(node.perm).toBe(0o600);
       expect(node.isFile()).toBe(true);
     });
-    describe.each([['uid'], ['gid'], ['atime'], ['mtime'], ['perm'], ['nlink']])('ctime changes', field => {
-      it(`set ${field}`, () => {
+    describe.each(['uid', 'gid', 'atime', 'mtime', 'perm', 'nlink'])('when %s changes', field => {
+      it('updates the property and the ctime', () => {
         const node = new Node(1);
         const oldCtime = node.ctime;
         node[field] = 1;
         const newCtime = node.ctime;
-        expect(newCtime !== oldCtime).toBeTruthy();
+        expect(newCtime).not.toBe(oldCtime);
+        expect(node[field]).toBe(1);
       });
     });
-
-    describe.each([
-      ['getString', []],
-      ['getBuffer', []],
-      ['read', [Buffer.alloc(0)]],
-    ])('atime changes', (method, args) => {
-      it(`${method}()`, () => {
+    describe('.getString(encoding?)', () => {
+      it('updates the atime and ctime', () => {
         const node = new Node(1);
         const oldAtime = node.atime;
         const oldCtime = node.ctime;
-        node[method](...args);
+        node.getString();
         const newAtime = node.atime;
         const newCtime = node.ctime;
-        expect(newAtime !== oldAtime).toBeTruthy();
-        expect(newCtime !== oldCtime).toBeTruthy();
+        expect(newAtime).not.toBe(oldAtime);
+        expect(newCtime).not.toBe(oldCtime);
+      });
+    });
+    describe('.getBuffer()', () => {
+      it('updates the atime and ctime', () => {
+        const node = new Node(1);
+        const oldAtime = node.atime;
+        const oldCtime = node.ctime;
+        node.getBuffer();
+        const newAtime = node.atime;
+        const newCtime = node.ctime;
+        expect(newAtime).not.toBe(oldAtime);
+        expect(newCtime).not.toBe(oldCtime);
       });
     });
   });
--- a/src/__tests__/promises.test.ts
+++ b/src/__tests__/promises.test.ts
@@ -462,7 +462,7 @@ describe('Promises API', () => {
       '/foo/baz': 'baz',
     });
     it('Read an existing directory', async () => {
-      expect(await promises.readdir('/foo')).toEqual(['bar', 'baz']);
+      expect(await promises.readdir('/foo')).toEqual(['.', '..', 'bar', 'baz']);
     });
     it('Reject when directory does not exist', () => {
       return expect(promises.readdir('/bar')).rejects.toBeInstanceOf(Error);
--- a/src/__tests__/promises.test.ts
+++ b/src/__tests__/promises.test.ts
@@ -462,7 +462,7 @@ describe('Promises API', () => {
       '/foo/baz': 'baz',
     });
     it('Read an existing directory', async () => {
-      expect(await promises.readdir('/foo')).toEqual(['.', '..', 'bar', 'baz']);
+      expect(await promises.readdir('/foo')).toEqual(['bar', 'baz']);
     });
     it('Reject when directory does not exist', () => {
       return expect(promises.readdir('/bar')).rejects.toBeInstanceOf(Error);
--- a/src/node.ts
+++ b/src/node.ts
@@ -44,9 +44,21 @@ export class Node extends EventEmitter {
     this.perm = perm;
     this.mode |= perm;
     this.ino = ino;
+
+    return new Proxy(this, {
+      set(target, name, value): boolean {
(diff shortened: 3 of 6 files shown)


src/__tests__/node.test.ts
```diff
@@ -69,5 +69,31 @@ describe('node.ts', () => {
       expect(node.perm).toBe(0o600);
       expect(node.isFile()).toBe(true);
     });
+    describe.each([['uid'], ['gid'], ['atime'], ['mtime'], ['perm'], ['nlink']])('ctime changes', field => {
+      it(`set ${field}`, () => {
+        const node = new Node(1);
+        const oldCtime = node.ctime;
+        node[field] = 1;
+        const newCtime = node.ctime;
+        expect(newCtime !== oldCtime).toBeTruthy();
```
G-Rath, line 78: Let's use `expect` matchers for these checks i.e. `expect(newCtime).not.toBe(oldCtime)`.

src/__tests__/promises.test.ts
```diff
@@ -462,7 +462,7 @@ describe('Promises API', () => {
       '/foo/baz': 'baz',
     });
     it('Read an existing directory', async () => {
-      expect(await promises.readdir('/foo')).toEqual(['bar', 'baz']);
+      expect(await promises.readdir('/foo')).toEqual(['.', '..', 'bar', 'baz']);
```
G-Rath, line 465: This behaviour is not correct, per the Node docs:

> Fulfills with an array of the names of the files in the directory excluding '.' and '..'.

Technically it doesn't say that for `readdirSync`, but does do that too:

```
~ via  v16.15.0 via 💎 v3.1.2
❯ mkdir empty

~ via  v16.15.0 via 💎 v3.1.2
❯ node
Welcome to Node.js v16.15.0.
Type ".help" for more information.
> fs.readdirSync('empty')
[]
>
```

G-Rath (changes requested):

src/__tests__/node.test.ts
vangie: done

src/__tests__/promises.test.ts
vangie: fixed

G-Rath:
@vangie would you mind applying prettier, then I think we can land this

vangie:
> @vangie would you mind applying prettier, then I think we can land this

@G-Rath fixed

G-Rath (approved):

G-Rath:
@vangie looks like it's still not happy

vangie:
> @vangie looks like it's still not happy, and the tests are all failing

@G-Rath Sorry, I tried to fix it again. I don’t know why the github action won’t be triggered immediately after I push it.

src/node.ts
```diff
@@ -44,9 +44,21 @@ export class Node extends EventEmitter {
     this.perm = perm;
     this.mode |= perm;
     this.ino = ino;
+
+    return new Proxy(this, {
```
G-Rath, line 48: Actually, why are we not using `get` & `set` ers for this? I think that would be better than returning a proxy

G-Rath:
[It's a security thing](https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/) - you should be able to run the tests locally though to ensure they work.

It looks like this is not compatible with #902 - @alienzhou might be able to help out here

src/node.ts
G-Rath: hmm I think we should keep things simple by `get` & `set` - Proxy tends to be more complex if things go wrong, and even less performant.

vangie: So is there any other suggested way to write it? something like this
```
['prop_1', 'prop_2' ... 'prop_n'].forEach(prop => {
      set [prop](value) {
        this[prop] = value;
      }
});
```

G-Rath: No, you just write it as a `get` and `set` in the class:

```
get uid(): number {
  return this._uid;
}

set uid(value: number) {
  this.ctime = new Date();
  this._uid = value;
}

// .. and so on for each property
```

vangie: OK

vangie: @G-Rath fixed

src/node.ts
```diff
@@ -19,34 +19,97 @@ export class Node extends EventEmitter {
   ino: number;
 
   // User ID and group ID.
-  uid: number = getuid();
-  gid: number = getgid();
+  _uid: number = getuid();
```
G-Rath, line 22: Could you add the appropriate modifiers to these properties + your new getters & setters?

src/__tests__/node.test.ts
```diff
@@ -69,5 +69,31 @@ describe('node.ts', () => {
       expect(node.perm).toBe(0o600);
       expect(node.isFile()).toBe(true);
     });
+    describe.each([['uid'], ['gid'], ['atime'], ['mtime'], ['perm'], ['nlink']])('ctime changes', field => {
```
G-Rath, line 72: You don't need these nested arrays:

```suggestion
    describe.each(['uid', 'gid', 'atime', 'mtime', 'perm', 'nlink'])('ctime changes', field => {
```

src/__tests__/node.test.ts
```diff
@@ -69,5 +69,31 @@ describe('node.ts', () => {
...
+        expect(newCtime).not.toBe(oldCtime);
+      });
+    });
+
+    describe.each([
+      ['getString', []],
+      ['getBuffer', []],
+      ['read', [Buffer.alloc(0)]],
+    ])('atime changes', (method, args) => {
+      it(`${method}()`, () => {
```
G-Rath, line 87: Since this doesn't use `field`, let's pull it up a level and change the title to be `when %s is called -> it updates the atime and ctime`

G-Rath (changes requested):
Looking good - just have a few tweaks I'd like to the test structure

src/node.ts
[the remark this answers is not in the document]
vangie: done

src/__tests__/node.test.ts
vangie: I have changed like this
```
    describe.each([
      ['getString', []],
      ['getBuffer', []],
      ['read', [Buffer.alloc(0)]],
    ])('when %s is called', (method, args) => {
      it(`updates the atime and ctime`, () => {
...
```

the `read` method cannot without the `Buffer` argument.

src/__tests__/node.test.ts
```diff
@@ -69,5 +69,31 @@ describe('node.ts', () => {
       expect(node.perm).toBe(0o600);
       expect(node.isFile()).toBe(true);
     });
+    describe.each([['uid'], ['gid'], ['atime'], ['mtime'], ['perm'], ['nlink']])('ctime changes', field => {
+      it(`set ${field}`, () => {
```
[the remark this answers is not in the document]
vangie: done

src/__tests__/node.test.ts
G-Rath: I'm not sure what you mean - to be clear by "pull it up a level" I mean do this:

```
describe.each(['uid', 'gid', 'atime', 'mtime', 'perm', 'nlink'])('when %s changes', field => {
 // ...
});

describe.each([
   ['getString', []],
   ['getBuffer', []],
   ['read', [Buffer.alloc(0)]],
 ])('when %s is called', (method, args) => {
   it(`updates the atime and ctime`, () => {
    // ...
  });
});
```

Personally I'm not the biggest fan of using the `describe.each` like this, but it's not the end of the world so if it works it works 🤷

vangie: Pull it up a level like this，I have try but if I remove the `read` method's argument `Buffer.alloc(0)`, the test cannot pass. So I don't know how to pull it up without removing `Buffer.alloc`
```
describe.each(['getString', 'getBuffer',  'read'])('when %s is called', method => {
   it(`updates the atime and ctime`, () => {
    // ...
  });
});
```

G-Rath: But why are you removing `Buffer.alloc(0)` at all? You shouldn't need to do that to pull it up

vangie: So how to get rid of `Buffer.alloc(0)`? Modify the `read` method, such as giving a default value. Or write a test for `read` alone instead of putting it in `describe.each`.

G-Rath: I'm not asking you to remove it, so you don't do anything with it 🙂

vangie: OK，I tried to change it, if you think it is inappropriate, please give more specific suggestions for modification

src/__tests__/node.test.ts
```diff
@@ -69,5 +69,39 @@ describe('node.ts', () => {
...
+        const oldCtime = node.ctime;
+        node[method]();
+        const newAtime = node.atime;
+        const newCtime = node.ctime;
+        expect(newAtime).not.toBe(oldAtime);
+        expect(newCtime).not.toBe(oldCtime);
+      });
+    });
+
+    it('when node.read is called, updates the atime and ctime', () => {
```
G-Rath, line 96: we've already got a `describe` for tests of `.read`, so can you move this under that? (and remove the `when ... called,` )

src/__tests__/node.test.ts
```diff
@@ -69,5 +69,39 @@ describe('node.ts', () => {
...
+        const node = new Node(1);
+        const oldCtime = node.ctime;
+        node[field] = 1;
+        const newCtime = node.ctime;
+        expect(newCtime).not.toBe(oldCtime);
+        expect(node[field]).toBe(1);
+      });
+    });
+
+    describe.each(['getString', 'getBuffer'])('when node.%s is called', method => {
```
G-Rath, line 83: I don't think its worth having an each for two methods like this - instead could you lay it out like this:

```
describe('.getString()', () => {
  it('updates the atime and ctime', () => {
    // ...
  });
});

describe('.getBuffer()', () => {
  it('updates the atime and ctime', () => {
    // ...
  });
});
```

This'll make it easier to add some more tests for those methods in future too.

G-Rath (changes requested):

src/__tests__/node.test.ts
vangie: OK

src/__tests__/node.test.ts
vangie: done

src/__tests__/node.test.ts
```diff
@@ -69,5 +69,51 @@ describe('node.ts', () => {
...
+        const oldAtime = node.atime;
+        const oldCtime = node.ctime;
+        node.getBuffer();
+        const newAtime = node.atime;
+        const newCtime = node.ctime;
+        expect(newAtime).not.toBe(oldAtime);
+        expect(newCtime).not.toBe(oldCtime);
+      });
+    });
+    describe('.read()', () => {
```
G-Rath, line 106: we've already got a `describe('.read(buf, off, len, pos)', ...` above on line 55 - could you move this test into that?

vangie: done

PR #893, opened 2023-01-04
fix: set `path` on errors when possible

Nokel81:
Signed-off-by: Sebastian Malton <sebastian@malton.name>


Nokel81:
According to Typescript the typing for an `ErrnoException` (which is what Node generally throws when doing IO activity) is the following:

```ts
interface ErrnoException extends Error {
      errno?: number | undefined;
      code?: string | undefined;
      path?: string | undefined;
      syscall?: string | undefined;
}
```

And it is useful to have the `path` as part of the error

G-Rath (approved):

PR #896, opened 2023-01-30
add volume.fromDirectory to service data migration

alexsunday:
for some special situation, we need data from exists filesystem, so, add a method on Volume.

### DIFF
--- a/demo/fromDirectory.ts
+++ b/demo/fromDirectory.ts
@@ -1,0 +1,9 @@
+import { Volume } from '../src/volume';
+
+const vol = new Volume();
+vol.fromDirectory('demo', '/app');
+
+const out = vol.toJSON();
+
+const v2 = Volume.fromJSON(out);
+console.log(v2.readdirSync('/app'));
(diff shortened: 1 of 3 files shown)


G-Rath (changes requested):
Can you please add some tests?

alexsunday:
> Can you please add some tests?
> 
> You also need to apply formatting

OK，added a case, and pretty the code.

G-Rath:
It doesn't look like you're attempting to copy any of the properties on the files - I feel like we should be doing that, though also I this is why we've generally not provided a utility method like this as people might want different behaviours depending on the use-case 🤔

alexsunday:
yes, file properties not copied, this is because there is no corresponding field for the type 'DirectoryJSON' in the code. I though that 'DirectoryJSON' was not prepared to save all the file attribute here?

I see that the example code on readme.md, section 'Usage' , the json struct is very simple, 
but sometimes when we want to express the files in exists filesystem, it will be more convenient if there is a method available.

if you  fell it's unnecessary, just close pr.

alexsunday:
please close this pr, use json store binary file is a wrong way.

alexsunday:
use json store binary file is a wrong way, closed.

G-Rath:
@alexsunday I'm not sure what you mean?

PR #909, opened 2023-06-06
fix: bump fs-monkey minimum version for rm support

### LINKED ISSUE #884: .rm/.rmSync are not available
image

It seems that memfs has implemented `.rm` and `.rmSync`:

https://github.com/streamich/memfs/blob/ce45b54317dcd38b7337ba1977aa74d836ae1d6d/src/volume.ts#L2044-L2053

But memfs uses [fs-monkey](https://github.com/streamich/fs-monkey) to bind methods:

https://github.com/streamich/memfs/blob/ce45b54317dcd38b7337ba1977aa74d836ae1d6d/src/index.ts#L39-L40

`.rm` and `.rmSync` are not supported in the `fs-monkey`'s latest version yet: https://github.com/streamich/fs-monkey/pull/362

Maybe updating this dependency will fix it.

blimmer:
Version 1.0.4 supports `.rm` and `.rmSync`: https://github.com/streamich/fs-monkey/releases/tag/v1.0.4

Technically, the existing version specification allows this drift: https://github.com/streamich/memfs/blob/09eab278ec0deb82dd4576dea27a48930e4c0e9b/package.json#L78-L80

However, for completeness, we should use `^1.0.4` so `rm*` "just works".

Closes #884

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     }
   },
   "dependencies": {
-    "fs-monkey": "^1.0.3"
+    "fs-monkey": "^1.0.4"
   },
   "devDependencies": {
     "@semantic-release/changelog": "^6.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2179,10 +2179,10 @@ fs-minipass@^2.0.0, fs-minipass@^2.1.0:
   dependencies:
     minipass "^3.0.0"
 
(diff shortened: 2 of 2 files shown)


G-Rath (approved):

blimmer:
Should be fixed now @G-Rath - thanks!

PR #910, opened 2023-06-14
File System Access API

streamich:
Implements  [File System Access API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_Access_API) on top of [Node FS API](https://nodejs.org/dist/latest-v20.x/docs/api/fs.html). I.e. one can instantiate a File System Access API on top of an `fs`-like API, the real `fs` module or the in-memory one from `memfs`.

This latest code is working only in Node v20+, as it is using `Blob`, `WritableStream` and other browser-first JS APIs. I don't want to polyfill those, as: (1) the other code still works on lower Node versions and is not affected; (2) Node v20 soon will be the "Active" one. So, essentially for CI testing I am thinking to run tests only on v18 and v20, and I will need to disable the latest tests on v18; and will completely remove v14, v16, and v17.

Also, some code requires to bump TypeScript compilation target to `es6` from `es5`, that is probably a good overdue change anyways.

This adds adapter from Node FS API to File System Access API, I might also add the other way around: from File System Access API to Node FS API.

Also, might be worth doing similar adapters for the [File and Directory Entries API](https://developer.mozilla.org/en-US/docs/Web/API/File_and_Directory_Entries_API).

One reason I would like to have these adapters (besides testing) is to be able to normalize all those file systems to Node's `fs` API, hence, a code that expects Node's `fs` would automatically (likely) work for File System Access API as well as File and Directory Entries API.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
streamich, line 3: This is needed to inherit from `WritableStream`, it is probably still possible to do it some other way. But I think we wanted to switch to es6 some time ago anyways.

G-Rath: Since this is a breaking change, I think we should create a `next` branch and land this change in its own PR to that - the new GitHub release workflow (which we've not yet switched to) supports doing prereleases so that'll let us test the release too.

Assuming you've setup the npm release token like you have for other repos, you just need to disable circleci so it won't try and do a double release when we merge to `master`

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+  test('can instantiate', () => {
+    const { dir } = setup();
+    expect(dir).toBeInstanceOf(NodeFileSystemDirectoryHandle);
+  });
+
+  describe('.keys()', () => {
+    test('returns an empty iterator for an empty directory', async () => {
+      const { dir } = setup();
+      const keys = dir.keys();
+      expect(await keys.next()).toEqual({ done: true, value: undefined });
```
G-Rath, line 23: nit: ideally we should be using `toStrictEqual`

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+    });
+
+    test('returns two folders', async () => {
+      const { dir } = setup({
+        folder: null,
+        'another/folder': null,
+      });
+      const list: string[] = [];
+      for await (const key of dir.keys()) list.push(key);
+      expect(list.length).toBe(2);
```
G-Rath, line 40:
```suggestion
      expect(list).toHaveLength(2);
```

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+      for await (const key of dir.keys()) list.push(key);
+      expect(list).toEqual(['file.txt']);
+    });
+  });
+
+  describe('.entries()', () => {
+    test('returns an empty iterator for an empty directory', async () => {
+      const { dir } = setup();
+      const keys = dir.entries();
+      expect(await keys.next()).toEqual({ done: true, value: undefined });
```
G-Rath, line 57: nit: using `await expect(...).resolves` gives some nicer output if the test fails:

```suggestion
      await expect(keys.next()).resolves.toStrictEqual({ done: true, value: undefined });
```

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+    });
+
+    test('returns two entries', async () => {
+      const { dir } = setup({
+        'index.html': '<nobr>Hello, world!</nobr>',
+        'another/folder': null,
+      });
+      const handles: NodeFileSystemHandle[] = [];
+      for await (const entry of dir.entries()) handles.push(entry[1]);
+      expect(handles.length).toBe(2);
```
G-Rath, line 89:
```suggestion
      expect(handles).toHaveLength(2);
```

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+      expect(handles.find(handle => handle.name === 'index.html')).toBeInstanceOf(NodeFileSystemFileHandle);
+      expect(handles.find(handle => handle.name === 'another')).toBeInstanceOf(NodeFileSystemDirectoryHandle);
+    });
+  });
+
+  describe('.values()', () => {
+    test('returns an empty iterator for an empty directory', async () => {
+      const { dir } = setup();
+      const values = dir.values();
+      expect(await values.next()).toEqual({ done: true, value: undefined });
```
G-Rath, line 99: nit: using `await expect(...).resolves` gives some nicer output if the test fails:

```suggestion
      await expect(values.next()).resolves.toStrictEqual({ done: true, value: undefined });
```

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+    });
+
+    test('returns two entries', async () => {
+      const { dir } = setup({
+        'index.html': '<nobr>Hello, world!</nobr>',
+        'another/folder': null,
+      });
+      const handles: NodeFileSystemHandle[] = [];
+      for await (const entry of dir.values()) handles.push(entry);
+      expect(handles.length).toBe(2);
```
G-Rath, line 129:
```suggestion
      expect(handles).toHaveLength(2);
```

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
```diff
@@ -0,0 +1,424 @@
...
+    });
+  });
+
+  describe('.getFileHandle()', () => {
+    test('throws "NotFoundError" DOMException if file not found', async () => {
+      const { dir } = setup({ a: null });
+      try {
+        await dir.getFileHandle('b');
+        throw new Error('Not this error.');
+      } catch (error) {
```
G-Rath, line 221: You should be able to use the `toThrow` matcher to handle these error tests nicer - i.e. `await expect(...).rejects.toThrow(new DOMException('<message>', '<name'>))`.

Among other things, Jest will fail the test if no error is thrown

src/node-to-fsa/__tests__/scenarios.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const fs = memfs({
+      '/tmp': null,
+      '/etc': null,
+      '/bin': null,
+      '/Users/kasper/Documents/shopping-list.txt': 'Milk, Eggs, Bread',
+    }) as IFsWithVolume;
+    const dir = nodeToFsa(fs, '/Users/kasper/Documents');
+    const shoppingListFile = await dir.getFileHandle('shopping-list.txt');
+    const shoppingList = await shoppingListFile.getFile();
+    expect(await shoppingList.text()).toBe('Milk, Eggs, Bread');
```
G-Rath, line 16:
```suggestion
    await expect(shoppingList.text()).resolves.toBe('Milk, Eggs, Bread');
```

src/node-to-fsa/__tests__/util.ts
```diff
@@ -0,0 +1,7 @@
+const nodeMajorVersion = +process.version.split('.')[0].slice(1);
+
+/**
+ * The `File` global is available only starting in Node v20. Hence we run the
+ * tests only in those versions.
+ */
+export const maybe = nodeMajorVersion >= 20 ? describe : describe.skip;
```
G-Rath, line 7: I think it would be better to call this something like `onlyOnNode20` so its clear under what condition the tests should run.

src/node-to-fsa/__tests__/NodeFileSystemFileHandle.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+      test('can truncate up', async () => {
+        const { dir, fs } = setup({
+          'file.txt': '012345',
+        });
+        const entry = await dir.getFileHandle('file.txt');
+        const writable = await entry.createWritable({ keepExistingData: true });
+        await writable.write({ type: 'truncate', size: 10 });
+        await writable.close();
+        expect(fs.readFileSync('/file.txt').length).toBe(10);
```
G-Rath, line 60:
```suggestion
        expect(fs.readFileSync('/file.txt')).toHaveLength(10);
```

G-Rath (changes requested):
This looks very cool! I've got some small changes I'd like to see (the main one is renaming the `maybe` utility), but nothing major.

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/scenarios.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/NodeFileSystemDirectoryHandle.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/NodeFileSystemFileHandle.test.ts
streamich: Yes, I know, but I don't like to rely too much on library specific (Jest in this case) assertions if that can be easily done with just code.

src/node-to-fsa/__tests__/util.ts
streamich: Done.

streamich:
Thanks you @G-Rath for the review! I've addressed your changes and added two more commits which further improve FSA (types and read/write mode). I'll create the `next` branch, with just build/CI/fixes changes, so we can use that to release v4 and test the new workflow.

G-Rath:
Awesome, sounds like a plan

PR #926, opened 2023-06-22
Release v4

streamich:
- In-memory `fs` changes
  - A number of in-memroy `fs` bugs were fixed.
  - Type interfaces of methods and options objects for in-memory `fs` were improved.
  - `mkdirp` and `mkdripSync` were removed, they were deprecated before. (You can pass the `"recursive"` flag to `mkdir` and `mkdirSync`, instead.)
  - Many re-usable Node.js `fs` utilities now live in the `/src/node` folder.
  - Promises API methods are now bound by default to their `this` object, just like in native `fs` module.
- File System Access API
  - Node `fs` to File System Access API adapter was added.
  - File System Access API to Node `fs` adapter was added.
    - In browser, support for synchronous `fs` methods was added through WebWorker.
- `crudfs` and `casfs`
  - `crudfs` implementation on top of Node `fs` was added.
  - `crudfs` implementation on top of File System Access API was added.
  - `casfs` implementation on top of `crudfs` was added.
- `print` utility was added which allows to recursively print a directory tree to terminal.
- `snapshot` utility was added which allows to create recursive binary snapshots of folders and then load them back into some folder.
- The `/demo` folder now contains multiple Webpack demos.
- Build changes and deprecations
  - TypeScript was upgraded from version 4 to 5.
  - `tslib` is now included as peer dependency, instead of TypeScript helpers being emitted into distributable.
  - TypeScript build target of the distributable is no `es2017`, instead of previously `es5`.
  - The library is now tested on Node v18+ versions.
  - `BigInt` shim is no longer shipped, `memfs` will rely on the `BigInt` provided by the JavaScript environment. You can shim that global, if necessary.
  - `fs-monkey` dependemcy was removed.
- New features in this major release are marked as `experimental`, which means those APIs are in preview, their public interface may have breaking changes even across minor releases.
- CircleCI was removed, now GitHub actions power all builds and releases.
- The `next` branch was tested, which deploys pre-releases.


G-Rath:
@streamich nothing springs to mind other than checking if we want to change the min. node version? Just as I'm not sure if we're completely compatible with Node 4 any more

streamich:
@G-Rath cool, will merge it on Monday then.

streamich:
I've bumped compile target to `es2017`, as that emits native async/await.

streamich:
Not sure how to test with which Node version `memfs` is compatible. It is tested on CI for v18, and v20. Maybe it is better to remove that engine setting altogether.

Regarding older versions, the only change that should require higher Node.js versions is the TypeScript build target change from `es5` to `es2017`. I don't think there were any other changes that would require higher Node.js versions for the existing code.

The new code added  in v4 release, is imported through a separate path; big part of it required the latest Node v20 to run; or a browser environment.

Issue #735, opened 2021-08-27
Missing features 😱 😣 😭 🤷‍♂️


Mesteery:
- [x] #1120
- [x] #1116
- [x] #1087
- [x] #1089
- [x] #1088
- [x] #663
- [x] #733
- [x] `readv` https://github.com/streamich/memfs/pull/946
- [x] `writev` https://github.com/streamich/memfs/pull/946
- [x] `readvSync` https://github.com/streamich/memfs/pull/946
- [x] `writevSync` https://github.com/streamich/memfs/pull/946
- [x] `rm` https://github.com/streamich/memfs/pull/754
- [x] `rmSync` https://github.com/streamich/memfs/pull/754
- [x] `promises.rm` https://github.com/streamich/memfs/pull/754
- [x] `throwIfNoEntry` in stats methods https://github.com/streamich/memfs/pull/782

Phillip9587:
@Mesteery Please add `fs.realpath.native` to the list. It is available since Node.js 9.2.0.
https://nodejs.org/api/fs.html#fsrealpathnativepath-options-callback

streamich:
https://github.com/streamich/memfs/issues/663

Nokel81:
FileHandle is missing `createReadStream` and `createWriteStream`

b1ek:
`fs.constants` is not implemented

streamich:
Recursive behavior of `readdir`: https://github.com/streamich/memfs/issues/967

SoulKa:
`Dirent.path` was deprecated and replaced by `Dirent.parentPath`. This is currently missing in memfs. I can make a PR if desired.

See [Dirent docs](https://nodejs.org/api/fs.html#class-fsdirent)

SoulKa:
Promises `writeFile()` does not support `Stream` as input, which it should. Issue is created at #1068

streamich:
All done!

PR #946, opened 2023-08-21
feat(volume): implement readv and writev

uhyo:
This PR addresses four missing features (plus two not mentioned there) listed in #735.

- Adds `readv`, `writev`, `readvSync` and `writevSync` to `Volume`.
- Adds `FileHandle#readv` and `FileHandle#writev`.

As a bonus, I modified the existing `read` implementation:
- modified the signature of `read` so that `position` can receive `null`, which better aligns to the `readv` signature. In fact, passing `null` was already working expect that TypeScript complained about that.
- passing `-1` to `position` is now equivalent to passing `null` for better compatibility with Node.
- wrote missing tests.
- implemented range check for the “read more than buffer space” case.

I also modified the existing `write` implementation:
- passing `-1` to `position` is now equivalent to passing `null` for better compatibility with Node.


G-Rath (approved):

uhyo:
@G-Rath Thank you for review! There was a conflict so I updated this PR. Hope you can recheck this 🙂

G-Rath (approved):

PR #952, opened 2023-09-18
feat(volume): `toJSON` now accepts the `asBuffer` parameter 

shorwood:
> *I had to recreate the PR since I fumbled a bit and force-pushed at some point*

## Context

Review feedback from the #880.
Note that this PR is isolated and does not include #880 

## Actions

Extended `vol.toJSON` so it can now output the virtual directory structure as a map of `Buffer` if the `asBuffer` parameter is set to `true`

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -524,7 +524,7 @@ export class Volume implements FsCallbackApi {
     });
   }
 
-  private _toJSON(link = this.root, json = {}, path?: string): DirectoryJSON {
+  private _toJSON(link = this.root, json = {}, path?: string, asBuffer?: boolean): DirectoryJSON {
     let isEmpty = true;
 
     let children = link.children;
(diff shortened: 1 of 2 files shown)


src/volume.ts
```diff
@@ -186,13 +186,13 @@ function validateGid(gid: number) {
 }
 
 // ---------------------------------------- Volume
-type DirectoryContent = string | null;
+type DirectoryContent = string | Buffer | null;
 
-export interface DirectoryJSON {
-  [key: string]: DirectoryContent;
+export interface DirectoryJSON<T extends DirectoryContent = DirectoryContent> {
```
shorwood, line 191: Allow type extension, this is because return type of `toJSON` and parameter type of `fromJSON` are now different. Using generic allows atomicity.

G-Rath (approved):

PR #960, opened 2023-10-23
docs: fix link to unionfs

regseb:
Fix link to `unionfs` in [_Usage_](https://github.com/streamich/memfs/blob/252a507/docs/node/usage.md#usage) page.


G-Rath (approved):

G-Rath:
I have force-pushed this onto `main` as https://github.com/streamich/memfs/commit/c8bfa38aa15f1d3c9f326e9c25c8972326193a26 because I missed it having `fix:` instead of `docs:` in the commit message, and we don't need a release for that; no ones fault, just me being too speedy with the button clicking + being use to the commit message being pulled from the title. 😅

PR #972, opened 2023-11-30
feat: add support for `recursive` to `readdir`

### LINKED ISSUE #967: readdir does not support "recursive" option
memfs does not support the "recursive" option on readdir. As a result it is unable to reproduce the correct result when using a volume with a nested tree.

Another related issue, when using the `withFileTypes: true` option the Dirent[] object returned by memfs does not contain the base `path` prop. I suspect this will be needed to properly implement the "recursive" option.

I am am testing code similar to the following, using memfs v4.6.0.

```js
import { readdir } from "node:fs/promises";
import { relative } from "node:path";

export async function getFilePaths({ basepath }) {
  const entries = await readdir(resolve(basepath), {
    recursive: true,
    withFileTypes: true,
  });
  return entries
    .filter((entry) => entry.isFile())
    .map((file) => relative(basepath, resolve(file.path, file.name)));
}
```

The test implemented with vitest.
```js
import { getFilePaths } from "./fileUtils.js";
import { vol } from "memfs";
import { afterEach, describe, test, expect, vi } from "vitest";

vi.mock("node:fs/promises", async () => {
  const memfs = await vi.importActual("memfs");
  return memfs.fs.promises;
});

describe("test", () => {
  afterEach(() => {
    vol.reset();
  });

  test('should return only files', async () => {
    // arrange
    const expected = "/r/37/oj33b0leub.md";
    vol.fromJSON({
      [expected]: "# kn7ay6b7hj\n1. isin7s29f6\n"
    }, "/");

    // act
    const actual = await getFilePaths({ basepath: "/r" });

    // assert
    expect(actual).toHaveLength(1);
    expect(actual[0]).toStrictEqual(expected);
  });
});

```

MichaelJCole:
Fix readdir recursive option and path to Dirent when withFileTypes: true

There wasn't a good way to do {recursive: true, withFileTypes: false} without making too many code paths.

I refactored it to calculate the data needed for { withFileTypes: true }, made a code path for function recursion if ( recursive: true }, and then original call returns just the paths if it's options were {withFileTypes: false} 

There may be a better way to do it, but that's what I got from my first look in the code.

Resolves #967

### DIFF
--- a/src/Dirent.ts
+++ b/src/Dirent.ts
@@ -15,11 +15,13 @@ export class Dirent implements IDirent {
 
     dirent.name = strToEncoding(link.getName(), encoding);
     dirent.mode = mode;
+    dirent.path = link.getPath()
 
     return dirent;
   }
 
   name: TDataOut = '';
+  path: TDataOut = '';
   private mode: number = 0;
 
   private _checkModeProperty(property: number): boolean {
--- a/src/Dirent.ts
+++ b/src/Dirent.ts
@@ -21,7 +21,7 @@ export class Dirent implements IDirent {
   }
 
   name: TDataOut = '';
-  path: TDataOut = '';
+  path = '';
   private mode: number = 0;
 
   private _checkModeProperty(property: number): boolean {
--- a/src/__tests__/volume/readdirSync.test.ts
+++ b/src/__tests__/volume/readdirSync.test.ts
@@ -73,30 +73,42 @@ describe('readdirSync()', () => {
   
   it('accepts option {recursive: true}', () => {
     const vol = create({
-      '/y/af': 'a',
-      '/y/b/bf': 'b',
-      '/y/c/c/cf': 'c',
+      '/y/af1': 'a',
+      '/y/af2': 'a',
+      '/y/b/bf1': 'b',
+      '/y/b/bf2': 'b',
+      '/y/c/c/.cf0': 'c',
+      '/y/c/c/cf1': 'c',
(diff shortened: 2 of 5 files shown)


MichaelJCole:
For https://github.com/streamich/memfs/issues/967

G-Rath:
@MichaelJCole it looks like your tests are a bit flakey on Windows - is that something you could look into for me?

I previously fixed this for [another test](https://github.com/streamich/memfs/pull/986) by just not being order dependent, since I don't think we're guaranteeing the order

MichaelJCole:
> @MichaelJCole it looks like your tests are a bit flakey on Windows - is that something you could look into for me?
> 
> I previously fixed this for [another test](https://github.com/streamich/memfs/pull/986) by just not being order dependent, since I don't think we're guaranteeing the order

Hey @G-Rath, I'm a linux nerd and don't have a windows machine to test.

This might be the line you're talking about:  https://github.com/streamich/memfs/pull/972/commits/b6005592ba119ffebd5916eabdf18dc4adfeeda8#diff-dde13e8c51a8a2e1bce9d7d5dc9ada7cb503492d5e43ab27b996fb6038dde296L1493

I don't know why it's there.  I copied it from another test to be on the team for windows :-)

Since the data isn't streamed,  Totally support your changes.

G-Rath:
@MichaelJCole all good - I managed to get a fix landed after debugging using the CI pipeline :)

PR #973, opened 2023-12-10
add optional "caseInsensitivity" for volume - to reproduce windows behaviour 

jburrow:
This was a PR introduced for version 3.2 - It required rebasing;

My PR just update this and resubmits it.

Original PR: https://github.com/streamich/memfs/pull/632 : Feb 9, 2021
https://github.com/ajafff/memfs/commit/b214f0ed5f17e0f9d147ade001237c3d99206d7a#diff-22b97ef1141174ca5a814b313942e79b626a55c7cd1ebe7b97d0756298b03641

NOTE:  
"test": "jest --maxWorkers 2", 

This is removed as it breaks debugging from vscode;


jmfirth-arkane:
We have a use case where we would like to see an option to support case insensitivity as well.  I would be happy to resolve any conflicts in this branch to see this PR to completion if it is a feature that would be accepted.

G-Rath:
@jmfirth-arkane go for it - aside from the unneeded formatting changes, I think the only change needed is lets use an `options` object instead of taking a bool as the second arg in case we decide to add other options in future

PR #977, opened 2024-01-08
fix: use internal buffer instead of global Buffer

jorenbroekema:
This prevents the need to shim Buffer on window in a browser based project, which reduces complexity in the build chain for such projects.


G-Rath:
@jorenbroekema would you mind rebasing this?

jorenbroekema:
@G-Rath done

PR #993, opened 2024-02-17
feat: remove unused methods and types within `json-joy`

G-Rath:


G-Rath:
Technically it could be argued this is breaking because we're at least exporting the encoder and decoder so someone in theory could use some of these methods - I'll revisit later to see if the reduction in size is worth the (probably very small) risk

PR #1005, opened 2024-02-21
fix: fix name validation regex to accept file names beginning with a period

nileflowers:
Hi,

I was using node-to-fsa and realized that files beginning with a period, as in hidden *nix files, throw 

```typescript
new TypeError(`Failed to execute '${method}' on '${klass}': Name is not allowed.`);
```
from `src/node-to-fsa/util.ts:assertName`.

I am assuming that this is not an intended behavior, if this is a correct assumption, please accept my pull request.

I modified the regex and added a regression test case.


G-Rath (approved):
thanks!

G-Rath:
(don't worry about that CI failure - it'll get "resolved" when the commits are squashed as part of merging)

PR #1031, opened 2024-04-30
Bump `@jsonjoy.com/util` package

streamich:


G-Rath:
fwiw we don't actually need to bump this explicitly since the new version is allowed by our package constraint

PR #1037, opened 2024-06-09
feat: adding optional file/path to node-to-fsa thrown DOMException to enhance error reporting

nileflowers:
… 

Hi,

I was using node-to-fsa and was getting `DOMException`s of different types, but it wasn't easy to figure out which files were related to these exceptions.

I added the file/path as a parameter to the exception constructors to assist with debugging and troubleshooting, and I modified test cases accordingly.


G-Rath:
Thanks - would you mind making this PR against https://github.com/streamich/fs-zoo instead though? we're in the process of extracting these out of `memfs` to lighten the package size for those that just want an in-memory fs like `webpack-dev-server`

nileflowers:
https://github.com/streamich/fs-zoo/pull/7

PR #1041, opened 2024-07-01
fix: ensure files in subdirectories are returned as buffers when calling `toJSON` with `asBuffer`

elegaanz:
Without that, `asBuffer` only works for files at the root of the volume, not those in sub-directories.


G-Rath:
Thanks, I think this make sense - any chance you could add a test though?

G-Rath (approved):
🥳

PR #1058, opened 2024-09-05
fix: add `parentPath` to `Dirent`

SoulKa:
# Changes

- Add missing `parentPath` property to `Dirent` which is equal to the `path` property

The `parentPath` property will replace the `path` property on `Dirent` in future node versions (see [Docs](https://nodejs.org/api/fs.html#class-fsdirent)).

References #735

### DIFF
--- a/src/Dirent.ts
+++ b/src/Dirent.ts
@@ -16,12 +16,14 @@ export class Dirent implements IDirent {
     dirent.name = strToEncoding(link.getName(), encoding);
     dirent.mode = mode;
     dirent.path = link.getParentPath();
+    dirent.parentPath = dirent.path;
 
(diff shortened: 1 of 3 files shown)


G-Rath (changes requested):
thanks - would you mind adding a test to assert that `parentPath` has the same value as `path`?

it'd also be great if you could double check that we're using `parentPath` everywhere ourselves (I'm not sure if we do actually use `Dirent#path`)

SoulKa:
I fixed the tests that were broken due to the new `parentPath` property. Is that enough? I also changed the one occurrence where `Dirent#path` was used.

G-Rath (approved):
thanks!

PR #1059, opened 2024-09-15
feat: improve permissions support

### LINKED ISSUE #1009: Permissions support
Hi! What level of support does memfs have for permissions today? 

I'm migrating from `mock-fs`, which allowed me to create read-only files in tests with something like

```js
mockFs({
  fileName: mockFs.file({ mode: 0o444 }); 
})
```

In `memfs`, it seems the read-only permissions aren't respected, i.e. this doesn't throw an error like I'd expect: 

```js
vol.fromJSON({
  fileName: 'some existing content',
});

vol.chmodSync('fileName', 0o444);
vol.writeFileSync('fileName', 'new content'); // doesn't throw
```

Should I expect this to work, or perhaps am I doing something incorrect?

Thanks!

BadIdeaException:
**Pursuant to our discussion in #1009, this PR adds basic permission support to memfs. It supports the POSIX-compliant r, w and x permissions for user, group and world. It does not support the more eclectic permissions such as SUID, GUID, or the sticky bit.**

Some remarks on design considerations/implementation details:

1. All permissions checking is done within `Volume`. `Node` and `Link` remain permission agnostic, to stay in keeping with how existence checking was done. To make this possible, tree walking had to be moved from `Link` to a [private method on the volume](https://github.com/BadIdeaException/memfs/blob/f80037f5a6f87966e84eb55a7aeffcf71352a4e8/src/volume.ts#L432), though. The overhead, as per your requirement, is pretty small.

2. Extensive tests are under `src/__tests__/volume/*`. I could not quite figure out by what logic tests go into `src/__tests__/volume.test.ts` as opposed to a dedicated file under `src/__tests__/volume/`. In the end, I decided the latter was probably sort of a `volume.test.d` directory.

3. I did not implement permissions checking for `chmod` and `chown`, because their security model largely depends on whether the user is root or not. (More specifically, on whether the executing process has certain capability flags set, which I'm not sure there is a way to check from within Node.js anyways.) (NB: For `chmod`, it also depends on whether the current user is the owner of the file to be changed.) For the most part, I left this out to prevent creating situations where the user locks themselves out from accessing a file, which is then irrevocable due to the lack of a concept of elevating privileges.

4. I think in the `open*` code path, there is duplicate symlink resolution: first by [openFile](https://github.com/BadIdeaException/memfs/blob/master/src/volume.ts#L735), then again by [openLink](https://github.com/BadIdeaException/memfs/blob/master/src/volume.ts#L692). I commented the second one out for a try, and all tests still passed. But since was already the case before I changed things around, I do not feel quite confident enough to delete it. In this PR, it is active code, but maybe you can take a look at it yourself.

Resolves #1009


G-Rath (changes requested):
this is looking good to me - would you mind just running `prettier` over the files? (don't worry about commitlint, as everything will get squashed when we merge 🙂)

BadIdeaException:
I found a bug in `utimes`, which required permissions on the target when it shouldn't have. Fixed.

G-Rath (approved):

PR #1065, opened 2024-10-01
feat: implement lutimes

BadIdeaException:
Added implementation of `lutimes` including tests. Permissions are respected.


BadIdeaException:
I don't know why Github insists on putting all the old (already merged) commits about permissions into this. I synced my fork thinking this should establish a new baseline, but apparently not.

G-Rath:
@BadIdeaException probably because you previously pushed your commits directly to your forks default branch rather than a dedicated branch, and I squashed when merging - so technically while all the _changes_ are in this repo, the _commits_ that you've got are not.

You should be able to resolve that by 
1. recreate the default branch in your fork
2. manually squash all the commits into one on your feature branch
3. rebase your feature branch onto your default branch

(2. is required to ensure conflicts as git will only consider each commit in isolation and our squashed commit represents the final product which might not be compatible with each individual commit - alternatively, you could do something like create a new branch from the updated default branch, cherry-pick your new commits, then force push to your existing branch)

PR #1066, opened 2024-10-02
feat: implement lutimes

BadIdeaException:
Let's try this again...

Continuation of https://github.com/streamich/memfs/pull/1065


BadIdeaException:
"Some checks were not successful"

This is a pre-existing test that failed due to timing out. This is not a test that I wrote, nor does it cover code that I wrote/changed. I think this is a fluke that was already present and probably only manifests sporadically.

G-Rath (approved):
aweome, thanks!

PR #1076, opened 2024-11-27
feat: implement `createReadStream` and `createWriteStream` on `FileHandle`

### LINKED ISSUE #1063: memfs.createWriteStream() exists, while filehandle.createWriteStream() doesn't
Hi Folks,
i noticed that #5 mentions that `memfs.createWriteStream()` works as intended.
however i was reading the official node documentation at https://nodejs.org/api/fs.html#filehandlecreatewritestreamoptions and apparently the same `createWriteStream()` can be called on a file descriptor generated by `fs.open()`.

while this is not a huge issue, are there any plans to align to the node API?
thank you.

graue:
Closes #1063

Thanks for making this library! My test needed these methods to exist so I took a stab at adding support.

### DIFF
--- a/src/node/FileHandle.ts
+++ b/src/node/FileHandle.ts
@@ -1,6 +1,6 @@
 import { promisify } from './util';
 import type * as opts from './types/options';
-import type { IFileHandle, IStats, TData, TDataOut, TMode, TTime } from './types/misc';
+import type { IFileHandle, IReadStream, IWriteStream, IStats, TData, TDataOut, TMode, TTime } from './types/misc';
 import type { FsCallbackApi } from './types';
 
 export class FileHandle implements IFileHandle {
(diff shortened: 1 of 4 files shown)


graue:
Well, that's embarrassing: the first version of this that I submitted didn't work, due to another issue: the `fs.createWriteStream` method as defined on memfs volumes did not allow the `fd` option to be a `FileHandle`, as [Node.js's `fs` module does](https://nodejs.org/api/fs.html#fscreatewritestreampath-options).

In my second commit, I've aligned that behavior with the Node API (which makes my first commit work), and in my third commit, I've added a test for this. Let me know if you want me to rebase, or handle any of this differently.

G-Rath:
> the fs.createWriteStream method as defined on memfs volumes did not allow the fd option to be a FileHandle

would you be ok doing a dedicated PR for that? it sounds like it's a clean bugfix that can get its own version and changelog entry

graue:
Done: #1077

I'll wait for that to be merged then rebase this.

graue:
Rebased!

src/node/FileHandle.ts
```diff
@@ -33,6 +33,14 @@ export class FileHandle implements IFileHandle {
     return promisify(this.fs, 'fdatasync')(this.fd);
   }
 
+  createReadStream(options: opts.IFileHandleReadStreamOptions): IReadStream {
```
G-Rath, line 36: would it be possible to get some basic tests for these?

graue: Done (adapted from the equivalent tests for the volume tests)

graue:
@G-Rath, I think this is ready to go! The X is only because it doesn't like my commit message for the second commit; squashing will fix that. Would be nice to get this in so I can use an official release in my project :)

G-Rath (approved):

PR #1077, opened 2024-11-28
fix: accept passing a `FileHandle` to `createReadStream` and `createWriteStream`

graue:
Letting this option be a `FileHandle` aligns the API with [Node.js's `fs` module](https://nodejs.org/api/fs.html#fscreatewritestreampath-options).

(Split out from #1076 per request)

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -1444,4 +1444,29 @@ describe('volume', () => {
       });
     });
   });
+  describe('.createReadStream', () => {
+    it('accepts filehandle as fd option', done => {
+      const vol = Volume.fromJSON({
+        '/test.txt': 'Hello',
+      });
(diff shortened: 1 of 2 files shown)


src/__tests__/volume.test.ts
```diff
@@ -1431,4 +1432,16 @@ describe('volume', () => {
       expect(new StatWatcher(vol).vol).toBe(vol);
     });
   });
+  describe('.createWriteStream', () => {
```
G-Rath, line 1435: any reason why you've added a test for just `createWriteStream`, but not `createReadStream`?

graue: just because they're very similar but sure I can write one for `createReadStream`

G-Rath (changes requested):
awesome, once you've applied Prettier to the test file, we should be good to go!

G-Rath (approved):

Issue #550, opened 2020-06-17
Auto-closing streams throw errors when the volume is cleared in the meantime


falkenhawk:
Hi,
I have a test which downloads a file with `supertest` and the file is served by setting the response body to read stream:
```js
ctx.body = fs.createReadStream(filePath);
```
Then after the test finishes, I have a hook where I reset the memfs volume:
```js
afterEach(() => {
  vol.reset();
});
```
It looks like when the stream is being destroyed, it asynchronously calls the method to close the file, but the volume is already reset at that point, and it throws the error:

```
  Error: EBADF: bad file descriptor, close
      at createError ((...)/node_modules/memfs/lib/volume.js:121:17)
      at Volume.getFileByFdOrThrow ((...)/node_modules/memfs/lib/volume.js:650:19)
      at Volume.closeSync ((...)/node_modules/memfs/lib/volume.js:878:25)
      at Immediate._onImmediate ((...)/node_modules/memfs/lib/volume.js:682:39)
      at processImmediate (internal/timers.js:456:21)
```

Can it be avoided somehow in this case, by adding a check somewhere along those lines if the volume was reset in the meantime? Or do you maybe have any suggestion for me to work around it?

streamich:
Could you await until the request completes in your test? So when `vol.reset()` is called the file is already closed.

streamich:
Another approach could be to not use `vol.reset()` but in each test create its own volume.

```js
test('...', async () => {
  const vol = new Volume();
  // ...
});
```

falkenhawk:
Thanks for all the suggestions,

> Could you await until the request completes in your test? So when `vol.reset()` is called the file is already closed.

I think I am already doing it... I am not sure here, but the `destroy` method of the memfs' readstream might be called asynchronously, after the request (and the test) is done 🤔

```js
// import request from supertest, set up server (koa instance) etc.
// ...

  it('uploads a file', async () => {
    const fileRes = await request(server)
      .get('/file/pixel.gif');

    const localFile = fs.readFileSync('./test/files/pixel.gif');
    assert.ok(localFile.equals(fileRes.body));
  });
```

and I prefer to use one global "volume" here, to be reused both in the test and "on the other side" of the request... 🤔

Greenheart:
I solved this by properly `await`ing all async test cases and let them finish before the test runner (`vitest` in my case) executed `vol.reset()` to clear the volume.

```ts
test('something', async () => {
  await doSomething()
})
```

[...]

PR #1081, opened 2024-12-21
fix: don't fail on closing fd after reset has been called (#550)

adiktofsugar:
Spent a day tracking this down so I figured I could put in a fix. To review, the issue is that if you have a stream that happens to be open when you call `reset`, you get an error like this (assuming jest, but I'm sure it comes as a surprise to anyone on any testing platform):
```
> npx jest
Error: Unhandled error. (Error: EBADF: bad file descriptor, close
    at createError (/path/to/project/node_modules/memfs/src/node/util.ts:146:17)
    at Volume.getFileByFdOrThrow (/path/to/project/node_modules/memfs/src/volume.ts:540:33)
    at Volume.closeSync (/path/to/project/node_modules/memfs/src/volume.ts:818:23)
    at Immediate._onImmediate (/path/to/project/node_modules/memfs/src/volume.ts:578:25)
    at processImmediate (node:internal/timers:478:21)
    at process.topLevelDomainCallback (node:domain:160:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24) {
  code: 'EBADF'
})
    at Volume.ReadStream.emit (node:events:507:17)
    at Volume.ReadStream.emit (node:domain:488:12)
    at /path/to/project/node_modules/memfs/src/volume.ts:2397:18
    at Immediate._onImmediate (/path/to/project/node_modules/memfs/src/volume.ts:580:9)
    at processImmediate (node:internal/timers:478:21)
    at process.topLevelDomainCallback (node:domain:160:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24) {
  code: 'ERR_UNHANDLED_ERROR',
  context: Error: EBADF: bad file descriptor, close
      at createError (/path/to/project/node_modules/memfs/src/node/util.ts:146:17)
      at Volume.getFileByFdOrThrow (/path/to/project/node_modules/memfs/src/volume.ts:540:33)
      at Volume.closeSync (/path/to/project/node_modules/memfs/src/volume.ts:818:23)
      at Immediate._onImmediate (/path/to/project/node_modules/memfs/src/volume.ts:578:25)
      at processImmediate (node:internal/timers:478:21)
      at process.topLevelDomainCallback (node:domain:160:15)
      at process.callbackTrampoline (node:internal/async_hooks:128:24) {
    code: 'EBADF'
  }
}
```
I find it difficult to imagine someone calling `reset` and expecting an error when the `setImmediate` callback fires, so that's where I made my change. Other ways of fixing I thought about were:
- add a `waitForFilesToClose`, but potentially you're testing something where they wouldn't close? also it requires an extra command that you'd only use for this purpose.
- if you could somehow "reset" the actual file system, I'd imagine it would abort all the ongoing operations as well, so this seems ideologically correct (to me)

### DIFF
--- a/src/__tests__/volume.test.ts
+++ b/src/__tests__/volume.test.ts
@@ -361,6 +361,23 @@ describe('volume', () => {
           '/good': 'bye',
         });
       });
+      it('Open streams should not be affected', async () => {
+        const vol = new Volume();
+        const json = {
+          '/hello': 'world',
+        };
+        vol.fromJSON(json);
+        await new Promise((resolve, reject) => {
+          vol
+            .createReadStream('/hello')
+            .on('data', () => null)
+            .on('close', resolve)
+            .on('end', () => {
+              vol.reset();
+            })
+            .on('error', reject);
+        });
+      });
     });
     describe('.openSync(path, flags[, mode])', () => {
       const vol = new Volume();
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -570,15 +570,9 @@ export class Volume implements FsCallbackApi, FsSynchronousApi {
   }
   */
 
-  private abortControllers: Set<AbortController> = new Set();
-
   private wrapAsync(method: (...args) => void, args: any[], callback: TCallback<any>) {
     validateCallback(callback);
-    const abortController = new AbortController();
-    this.abortControllers.add(abortController);
     setImmediate(() => {
-      this.abortControllers.delete(abortController);
-      if (abortController.signal.aborted) return;
       let result;
       try {
         result = method.apply(this, args);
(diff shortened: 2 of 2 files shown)


G-Rath:
Can you confirm that this doesn't happen with the native fs? And if so, could you add a test too?

adiktofsugar:
@G-Rath you can't reset the native fs, so I'm not sure what I'd be comparing to. The closest I can think of would be if the node process closed while a stream is open, but node keeps itself open in that case unless it detects a cycle. I can definitely add a test, though.

adiktofsugar:
I thought maybe "reset" could be the same as "deleting all the files", so I tried to see what would happen if I deleted an open file and then ran close.:
```
const fd = fs.openSync(file);
fs.unlinkSync(file);
fs.closeSync(fd);
```
It doesn't throw an error. If you close it twice, though, it does (there's a test for that). So I decided to just do the `getFileByFdOrThrow(fd, 'close');` step in the `close` function (synchronously) and then call `closeFile()` directly rather than `closeSync`. I think that's the most correct fix, but let me know if you don't agree.

G-Rath (approved):
thanks - I didn't realise `reset` was our own extra function rather than a Node one 🙂

PR #1082, opened 2024-12-23
fix: allow setting custom file types beyond S_IFREG and S_IFDIR

kylecarbs:
Prior to this change, the `mode` argument to `Node` and `Volume` was only used to set the file permissions.

### DIFF
--- a/src/__tests__/node.test.ts
+++ b/src/__tests__/node.test.ts
@@ -3,9 +3,9 @@ import { constants } from '../constants';
 
 describe('node.ts', () => {
   describe('Node', () => {
-    const node = new Node(1);
+    const node = new Node(1, constants.S_IFREG | 0o666);
     it('properly sets mode with permission respected', () => {
-      const node = new Node(1, 0o755);
+      const node = new Node(1, constants.S_IFREG | 0o755);
       expect(node.perm).toBe(0o755);
       expect(node.mode).toBe(constants.S_IFREG | 0o755);
       expect(node.isFile()).toBe(true); // Make sure we still know it's a file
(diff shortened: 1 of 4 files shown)


kylecarbs:
Fixed up the formatting!

kylecarbs:
@G-Rath if you could take a look at this when you get a sec, that'd be awesome!

src/__tests__/node.test.ts
```diff
@@ -72,16 +72,16 @@ describe('node.ts', () => {
...
-      const node = new Node(1);
+      const node = new Node(1, constants.S_IFREG | 0o666);
       expect(node.perm).toBe(0o666);
       expect(node.isFile()).toBe(true);
       node.chmod(0o600);
       expect(node.perm).toBe(0o600);
       expect(node.isFile()).toBe(true);
     });
-    describe.each(['uid', 'gid', 'atime', 'mtime', 'perm', 'nlink'])('when %s changes', field => {
+    describe.each(['uid', 'gid', 'atime', 'mtime', 'nlink'])('when %s changes', field => {
```
G-Rath, line 82: question: why has `perm` been removed? shouldn't updating the `perm` still update the `ctime`?

src/node.ts
```diff
@@ -29,20 +29,17 @@ export class Node extends EventEmitter {
...
+  mode: number; // S_IFDIR, S_IFREG, etc..
 
   // Number of hard links pointing at this Node.
   private _nlink = 1;
 
   // Path to another node, if this is a symlink.
   symlink: string;
 
-  constructor(ino: number, perm: number = 0o666) {
+  constructor(ino: number, mode: number) {
```
G-Rath, line 40: question: is it possible to retain this default?

technically this is a public API, though I don't think we ever document it explicitly - but if we can retain the default value that'll make this slightly less breaking-change-y

src/volume.ts
```diff
@@ -359,7 +359,15 @@ export class Volume implements FsCallbackApi, FsSynchronousApi {
       throw new Error('createLink: name cannot be empty');
     }
 
-    return parent.createChild(name, this.createNode(isDirectory, perm));
+    // If no explicit permission is provided, use defaults based on type
+    const defaultPerm = isDirectory ? 0o777 : 0o666;
+    const finalPerm = mode ?? defaultPerm;
```
G-Rath, line 364: nit: I don't mind this being inlined since it's the only place we use the variable, and we've got a comment that explicitly calls out that we're got a default so I don't think it'll be easy to misunderstand

```suggestion
    const finalPerm = mode ?? (isDirectory ? 0o777 : 0o666);
```

kylecarbs:
@G-Rath I appreciate the good review! All fixes have been done, and I made the `Node` default to the old mode for backward compatibility.

src/__tests__/node.test.ts
```diff
@@ -18,33 +18,33 @@ describe('node.ts', () => {
     });
     describe('.write(buf, off, len, pos)', () => {
       it('Simple write into empty node', () => {
-        const node = new Node(1);
+        const node = new Node(1, 0);
```
G-Rath, line 21: now that we've restored the default, do we need to be passing in this explicit value? I feel like since this is the only change in these tests, it would be better if we didn't change them at all as we're expecting the current behaviour to not change right?

I'm happy if you want to add more tests with different modes/perms, if you think that would be useful

kylecarbs: Good call. Fixed!

Testing various scenarios would mostly be redundant. I _think_ current chardev test should catch any future bugs!

G-Rath (approved):

Issue #733, opened 2021-08-27
readSync and others functions doesn't really supports uint8array


Mesteery:
Functions such as `readSync` do not really support Uint8Array. It would be nice to support a Uint8Array, or better yet, like Node.js, all Typed Array and DataView.

corwin-of-amber:
In a browser environment, and with `buffer@4.9.2`, it works by coincidence. But with newer versions of `buffer`, it generates this error:
```js
> fd = mfs.openSync('/proj1/foo', 'r')
> mfs.readSync(fd, new Uint8Array(500), 0, 500)
Uncaught TypeError: argument should be a Buffer
    at Uint8Array.copy (index-13.js:1691)
    at Node.read (node-6.js:122)
    at File.read (node-6.js:380)
    at Volume.readBase (volume-3.js:878)
    at Volume.readSync (volume-3.js:882)
    at <anonymous>:1:5
```

I encountered the problem while using [wasmer-js](https://github.com/wasmerio/wasmer-js).

mattfysh:
any chance this might have regressed? ive added a monkey patch

```
fs.readSync = (fd: number, buf: Uint8Array, off: number, len: number, pos: number | null) => {
  const nodeBuf = Buffer.isBuffer(buf)
    ? (buf as unknown as Buffer)
    : Buffer.from(buf.buffer, buf.byteOffset, buf.byteLength);
  const n = _readSync(fd, nodeBuf, off, len, pos);
  if (!Buffer.isBuffer(buf)) {
    buf.set(nodeBuf.subarray(off, off + n), off);
  }
  return n;
};
```

mattfysh:
opened pull request: https://github.com/streamich/memfs/pull/1184

PR #1083, opened 2024-12-31
feat: support `UInt8Array` in place of `Buffer`

kylecarbs:
This seems reasonable, considering memfs supports the browser, and users are more likely to have instances of Uint8Array there.

Resolves https://github.com/streamich/memfs/issues/733

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -58,9 +58,9 @@ import {
   getWriteSyncArgs,
   unixify,
 } from './node/util';
-import type { PathLike, symlink } from 'fs';
+import type { symlink } from 'fs';
+import type { PathLike } from './node/types/misc';
 import type { FsPromisesApi, FsSynchronousApi } from './node/types';
-import { fsSynchronousApiList } from './node/lists/fsSynchronousApiList';
 import { Dir } from './Dir';
 
 const resolveCrossPlatform = pathModule.resolve;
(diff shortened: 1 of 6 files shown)


G-Rath:
I feel like this probably is a fix for #733 - would you mind double checking that for me, and throwing in a test or two as/when you think is needed? (if you don't think it's worth having any, I'm ok with that too 🙂)

kylecarbs:
@G-Rath seems to fix #733! I added tests before, but they don't do much - the structure is the same for Uint8Array. I'm happy to add it if you'd prefer, though!

G-Rath:
Awesome! Mind sticking a "resolves" line to your description?

Re tests it doesn't surprise me but would still be good if we had at least one where were explicitly passing in a 8uintarry

kylecarbs:
@G-Rath added to the description! I can add one before merge - no biggie.

kylecarbs:
@G-Rath fixed. I also removed a `debugger;` line that was causing me some issues, just assuming that's fine.

kylecarbs:
Seems like there are some unrelated flakes failing - but test I added should be good.

G-Rath:
>  I also removed a debugger;

Yeah that should never have been landed - I've opened #1086 to remove it and another from a test, just so we've got a dedicated changelog entry as it's a whole bug of its own

> Seems like there are some unrelated flakes failing

A couple of the watch ones are very slightly flakey, but thankfully only slightly so yes after rerunning you've got green across the board 🎉

src/volume.ts
```diff
@@ -58,9 +58,9 @@ import {
   getWriteSyncArgs,
   unixify,
 } from './node/util';
-import type { PathLike, symlink } from 'fs';
+import type { symlink } from 'fs';
+import type { PathLike } from './node/types/misc';
```
G-Rath, line 62: we're re-exporting `symlink` so we can merge these

```suggestion
import type { PathLike, symlink } from './node/types/misc';
```

G-Rath (approved):

PR #1085, opened 2025-01-09
feat: allow setting rdev on node

kylecarbs:
This adds it to stat too, so rdev will propagate.

### DIFF
--- a/src/Stats.ts
+++ b/src/Stats.ts
@@ -25,7 +25,7 @@ export class Stats<T = TStatNumber> {
     stats.uid = getStatNumber(uid);
     stats.gid = getStatNumber(gid);
 
-    stats.rdev = getStatNumber(0);
+    stats.rdev = getStatNumber(node.rdev);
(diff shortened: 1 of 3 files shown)


G-Rath (changes requested):
could you add a test or two for this? iiuc this should mean `fs.stat` can now return a value other than zero, so we should be able to have some tests showing that off

kylecarbs:
True. Will change.

kylecarbs:
@G-Rath this gets a bit awkward because it would be made with the `mknod` syscall, which is unavailable in NodeJS.

I can add a test inside of `Node` at least. I'm just using it internally anyways, so that works.

kylecarbs:
@G-Rath I added a test to `statSync` which seemed like the best spot.

G-Rath (approved):

kylecarbs:
@G-Rath fixed the prettier issue!

G-Rath (approved):

PR #1100, opened 2025-05-02
perf: use promises rather than `setImmediate` so that I/O is executed within the current task stack

VisualSJ:
If setImmediate is used, when there are too many setTimeout, IO will be delayed in execution, and this execution will cause the IO operation interval to become extremely long.
Referring to the NodeJS source code, I/O operations should use microtasks


G-Rath:
Is there any kind of test or samples you can provide on this? I don't have a big problem with this as is since it gets rid of a file, but I'd like to understand a bit more about the background and (scale of) impact if possible

VisualSJ:
Inside wrapAsync, it is implemented using setImmediate. In this case, it will create a new macro task at the end of the JavaScript task queue. The actual I/O operation will only commence when the macro task is executed.

That is to say, when the user initiates a readFile request, it will be inserted at the end of the JavaScript task queue. The actual I/O operation will only start after the current JavaScript task queue has been completely executed.

When using microtasks (Promise), it is equivalent to inserting a high-priority task into the task queue. It will be executed immediately at the end of the current macro task instead of being placed at the end of all macro tasks.

When there is a for loop that executes readFile within the loop and there are a large number of setTimeout calls in the program, the execution speed will become very slow.

VisualSJ:
test.a.js
```javascript
function wrapA(callback) {
  setImmediate(() => {
    callback();
  });
}
function readFileA(file, callback) {
  wrapA(() => {
    // nothing
    callback();
  });
}

(async () => {
  setTimeout(() => {
    const date = Date.now();
    while (Date.now() - date < 1000) {
      // nothing
    }
  });
  console.time('A');
  for (let i = 0; i < 1000; i++) {
    await new Promise(resolve => {
      readFileA('a.txt', resolve);
    });
  }
  console.timeEnd('A');
})();

```

test.b.js
```javascript
function wrapB(callback) {
  Promise.resolve().then(() => {
    callback();
  });
}
function readFileB(file, callback) {
  wrapB(() => {
    // nothing
    callback();
  });
}

(async () => {
  setTimeout(() => {
    const date = Date.now();
    while (Date.now() - date < 1000) {
      // nothing
    }
  });
  console.time('B');
  for (let i = 0; i < 1000; i++) {
    await new Promise(resolve => {
      readFileB('a.txt', resolve);
    });
  }
  console.timeEnd('B');
})();

```

VisualSJ:
@G-Rath 
Do I need to add any more information?
I changed the PR name according to the rules, but it seems that CI didn't recheck? Could you help trigger this manually and give it a try? Or could you tell me how to change the PR name? 😅

G-Rath (approved):
you're all good, I just hadn't gotten back around to this - I have updated the title to reflect this is a performance change and describe better what the actual change was, but that's just a nit thing :)

PR #1216, opened 2025-11-27
fix: add glob functions to API lists

DMartens:
**Problem**:
The API lists seem not to have been updated when the `glob(Sync)` methods were added.
As a result `fs` instances created with `createFsFromVolume` do not have the `glob` methods available.

**Solution**:
This PR adds `glob` to the callback API list and `globSync` to sync API list.


streamich (approved):
Thank you!

Issue #23, opened 2017-08-29
Doesn’t work with relative paths


sapegin:
After updating to 2.x relative paths don’t work:

```
> const { fs } = require('memfs');
undefined
> fs.writeFileSync('f', 't');
TypeError: Cannot read property 'createChild' of null
    at Volume.createLink (/Users/sapegin/Dropbox/Projects/_Repos/mrm-core/node_modules/memfs/lib/volume.js:423:22)
    at Volume.openFile (/Users/sapegin/Dropbox/Projects/_Repos/mrm-core/node_modules/memfs/lib/volume.js:690:29)
    at Volume.openBase (/Users/sapegin/Dropbox/Projects/_Repos/mrm-core/node_modules/memfs/lib/volume.js:698:25)
    at Volume.writeFileBase (/Users/sapegin/Dropbox/Projects/_Repos/mrm-core/node_modules/memfs/lib/volume.js:927:23)
    at Volume.writeFileSync (/Users/sapegin/Dropbox/Projects/_Repos/mrm-core/node_modules/memfs/lib/volume.js:952:14)
    at repl:1:4
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
> fs.writeFileSync('/f', 't');
undefined
```

sapegin:
Trying to debug the issue: it tries to resolve the file path using my current working directory with is obviously doesn’t exist in memfs volume:

```
> const { fs } = require('memfs');
undefined
> process.cwd()
'/Users/sapegin/Dropbox/Projects/_Repos/mrm-core'
> process.chdir('/');
undefined
> fs.writeFileSync('f', 't');
undefined
```

streamich:
How do you expect it to work with relative paths?

sapegin:
It used to work in 0.x so I’m not sure what has changed. It’s either a regression or should be documented that `process.chdir('/');` is required to work with relative paths.

streamich:
> It used to work in 0.x 

How can I reproduce it? Because if I install version `0.0.15` and try to `writeFileSync('f', 't')` it also throws an error, because it also depends on current working directory.

>  I’m not sure what has changed

The version `2.x` is a complete rewrite, with support for `Buffer`, symlinks, hardlinks etc.

> should be documented that process.chdir('/'); is required to work with relative paths

[...]

PR #1224, opened 2025-12-18
feat: partial relative paths support

Mopsgamer:
> [!WARNING]
> There was also a useless line nearby. It might be worth submitting it independently of this PR.

Refs #23

__Partial support__ and not fully tested. Can this be merged as is, or should contributors finalize it?

## Features
- New `cwd` property (getter + setter) is available for `Volume` and `Superblock` instances.

### DIFF
--- a/docs/node/relative-paths.md
+++ b/docs/node/relative-paths.md
@@ -30,5 +30,6 @@ PR [1124](https://github.com/streamich/memfs/pull/1224) adds partial support for
 relative paths. Tests are passing on the following methods:
 
 - `fs.promises.readFile`
+- `fs.readFileSync`
 - `vol.readdirSync`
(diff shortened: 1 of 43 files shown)


Mopsgamer:
Sorry, I missed the merge conflicts. It was quite difficult to sort through the new changes. I noticed there are two options involving `cwd`, but it seems this PR needs another one.

The current interface looks like this:
`memfs(json, cwd1, { process: {cwd2()}})`

Right now, `cwd2` acts as the mount point. From what I understand, `cwd1` serves as the mount for `cwd2`, which seems a bit counterintuitive.

We should either add a new `cwd3` option or redefine how `cwd1` behaves.

`cwd3` would set the current directory for resolving relative file paths within the methods.

PR #1253, opened 2026-03-26
fix(access): ensure symlinks are followed when checking for executable flag

### LINKED ISSUE #1252: access/accessSync w/ X_OK does not follow symlinks
`fs.accessSync()` with `X_OK` checks the symlink node's own mode instead of following the symlink to check the target's permissions. This diverges from POSIX [access(2)](https://man7.org/linux/man-pages/man2/access.2.html) behavior, which is to follow symlinks by default.

`R_OK` and `F_OK` work correctly through symlinks; only `X_OK` is affected.

## Example Reproduction

```js
import {memfs} from 'memfs';
const {vol, fs} = memfs();

vol.fromJSON({'/test.txt': 'content'});
vol.chmodSync('/test.txt', 0o755);
vol.symlinkSync('/test.txt', '/test-link');

// Target is executable — works fine
fs.accessSync('/test.txt', fs.constants.X_OK); // OK

// Through symlink — should also work, but throws EACCES
fs.accessSync('/test-link', fs.constants.X_OK); // EACCES

// Other modes work fine through the symlink
fs.accessSync('/test-link', fs.constants.R_OK); // OK
fs.accessSync('/test-link', fs.constants.F_OK); // OK
```

I've verified this also occurs with `fs.access()` and `fs/promises.access()`.

## Expected Behavior

`accessSync('/test-link', X_OK)` should follow the symlink and check execute permission on the target (`/test.txt`, mode `0o755`), returning successfully.

## Env

- memfs: v4.57.1
- Node.js: v24.14.0
- macOS Tahoe

## Notes

The symlink node is created with a default mode of `0o666`. For `R_OK` and `F_OK`, the symlink is followed correctly, but for `X_OK` the permission check _seems like_ it's running against the symlink node itself.

This may be related to #1250.

boneskull:
This is a one-line change in `volume.ts` which always resolves paths in the inner `_access` implementation.

The result is that executable symlinks are now detected properly by `access()`, `accessSync()` and `promises.access()`.

Added test suites for each of APIs asserting as much.

Fixes #1252

### DIFF
--- a/packages/fs-node/src/__tests__/volume/access.test.ts
+++ b/packages/fs-node/src/__tests__/volume/access.test.ts
@@ -1,0 +1,104 @@
+import { create } from '../util';
+import { AMODE } from '@jsonjoy.com/fs-node-utils';
+
+const setup = () => {
+  const vol = create({ '/test.txt': 'content' });
+  vol.chmodSync('/test.txt', 0o755);
+  vol.symlinkSync('/test.txt', '/test-link');
+  return vol;
+};
+
+describe('access', () => {
+  test('F_OK follows symlink', done => {
+    const vol = setup();
+    vol.access('/test-link', AMODE.F_OK, err => {
+      try {
+        expect(err).toBeNull();
+        done();
+      } catch (failure) {
+        done(failure);
+      }
+    });
+  });
+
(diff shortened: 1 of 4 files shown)


streamich (approved):
Thank you!

PR #1256, opened 2026-04-17
fix: uncomment cpSync and statfsSync in fsSynchronousApiList

GoodbyeNJN:
https://github.com/streamich/memfs/blob/74c356096bd1458675fd2028d2b07d1f2de6db67/packages/memfs/src/index.ts#L40-L42

The `createFsFromVolume` function relies on this list to add various methods to the `fs` object, and both `cpSync` and `statfsSync` have been implemented; they should be uncommented to be properly added to the `fs` object.


streamich (approved):
Thank you!

Issue #942, opened 2023-08-18
Discrepancy in close-after-unlink behavior


dmurvihill:
The `fs` module accepts calls to `FileHandle.close()` when the underlying file no longer exists, while `memfs` throws `ENOENT`. I suspect this could be the root cause of #550 and #900.

Reproduction cases (ts-jest):

```typescript  
describe('memfs/fs parity', () => {
  const testPath = path.join(__dirname, 'my-test-file');
    
  it('fs does not error on close after delete', async () => {
    const f = await fs.open(testPath, 'w');
    await fs.unlink(testPath);
    await f.close();
  });

  it('memfs should not error on close after delete', async () => {
    const f = await memfs.fs.promises.open(testPath, 'w');
    await memfs.fs.promises.unlink(testPath);
    await f.close();
  });
});
```

Error message:
```
 FAIL  test/fs.spec.ts (5.259 s)
  ● memfs/fs parity › memfs should not error on close after delete

    ENOENT: no such file or directory, open '/Users/dolan/IdeaProjects/big-shuffle/test/my-test-file'

      at createError (node_modules/memfs/src/node/util.ts:138:17)
      at Volume.openFile (node_modules/memfs/src/volume.ts:684:38)
      at Volume.openBase (node_modules/memfs/src/volume.ts:696:23)
      at Immediate.<anonymous> (node_modules/memfs/src/volume.ts:516:25)
```

G-Rath:
Makes sense - happy to accept a PR for that

PR #1261, opened 2026-06-20
fix: clamp negative truncate length to zero (uninitialized-memory leak)

chatman-media:
## Problem

`Node.truncate(len)` (in `fs-core`) only special-cases `len === 0`. A **negative** length falls through to the `if (len <= this.size)` branch and sets `this.size` to the negative value:

```ts
truncate(len: number = 0) {
  if (!len) { /* empty file */ return; }   // only len === 0
  if (len <= this.size) this.size = len;    // -1 <= 5  →  size = -1
  ...
}
```

`getBuffer()` then does `this.buf.subarray(0, this.size)`. With `this.size === -1`, `subarray` treats the end index relative to the buffer's length and returns bytes **past the used region** — i.e. uninitialized memory from the `bufferAllocUnsafe` backing buffer.

### Reproduction

```ts
import { Volume } from 'memfs';

const vol = new Volume();
vol.writeFileSync('/a', 'hello');
vol.truncateSync('/a', -1);

vol.readFileSync('/a').length;            // 63  (!!)  — expected 0
vol.readFileSync('/a').toString('hex');   // 68656c6c6f<uninitialized heap bytes…>
```

Node.js for comparison:

```js
fs.truncateSync(file, -1);   // file is now 0 bytes
fs.ftruncateSync(fd, -100);  // 0 bytes
```

Node's `fs.truncate` / `fs.ftruncate` clamp a negative length to an empty file (the C++ layer floors negative offsets at 0). memfs instead produced a *larger* file leaking adjacent heap contents.

## Fix

Treat a negative `len` the same as `0` in `Node.truncate` — change the `len === 0` guard to `len <= 0`. One-line behavioral change; `len === 0` is unaffected.

## Tests

Regression tests added at two levels, both verified to **fail before** the fix and **pass after**:

- `fs-core` unit: `Node#truncate(-1)` → size `0`, empty buffer.
- `fs-node` public API: `vol.truncateSync('/neg', -1)` → `readFileSync(...).length === 0` (guards against the leak).

Full `fs-core` + `fs-node` suites pass (817 tests), `typecheck` and `prettier:check` clean.

## Context

While investigating #942 (close-after-unlink — already resolved by the `Superblock` refactor), differential testing against Node's real `fs` surfaced this separate negative-length truncate parity gap, which has the added concern of exposing uninitialized memory.

### DIFF
--- a/packages/fs-core/src/Node.ts
+++ b/packages/fs-core/src/Node.ts
@@ -222,7 +222,12 @@ export class Node {
   }
 
   truncate(len: number = 0) {
-    if (!len) {
+    // A negative length is clamped to zero, matching Node.js `fs.truncate`,
+    // which truncates to an empty file rather than producing a negative size.
+    // Without this, `this.size` becomes negative and `getBuffer()`'s
+    // `subarray(0, this.size)` reads past the used region, exposing
+    // uninitialized heap memory from the unsafe-allocated backing buffer.
+    if (len <= 0) {
       this.buf = EMPTY_BUFFER;
       this.capacity = 0;
       this.size = 0;
--- a/packages/fs-core/src/__tests__/Node.test.ts
+++ b/packages/fs-core/src/__tests__/Node.test.ts
@@ -72,6 +72,18 @@ describe('Node', () => {
       expect(node.getBuffer().length).toBe(0);
     });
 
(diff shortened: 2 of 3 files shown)


streamich (approved):
Thank you!

PR #1269, opened 2026-07-15
chore: remove tslib from peerDependencies

### LINKED ISSUE #1241: tslib listed as peer and production dependency in memfs package.json
I'm in the process of updating some dependencies and memfs is a transitive dependency on one of my project's dependencies. When I updated my dependency, the current version of memfs (4.56.10) was added and pnpm started reporting peer dependency warnings due to `tslib`.

```bash
 ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies

apps/app.admin
└─┬ @rspack/cli 1.7.5
  └─┬ @rspack/dev-server 1.1.5
    └─┬ webpack-dev-server 5.2.2
      └─┬ webpack-dev-middleware 7.4.2
        └─┬ memfs 4.56.10
          ├── ✕ missing peer tslib@2
<snip>
Peer dependencies that should be installed:
  tslib@">=2.0.0 <3.0.0-0"
```

Looking at the `package.json` for the memfs package it looks a little off. `tslib` is listed as a [peer dependency](https://github.com/streamich/memfs/blob/54c7cce612b92075e853defe4c622494d06080df/packages/memfs/package.json#L127) with a version spec of `"2"` and also as a [production dependency](https://github.com/streamich/memfs/blob/54c7cce612b92075e853defe4c622494d06080df/packages/memfs/package.json#L143C5-L143C22) with a version spec of `"^2.0.0"`. 

It's possible that pnpm shouldn't report an unmet peer when the dependency specifying the peer also specifies it as a dependency. However, it seems unintentional for memfs to have `tslib` as both a dependency and a peer dependency as I can't think of a reason why you would need it to be both.

xianjianlf2:
## Summary

Fixes #1241

`packages/memfs/package.json` lists `tslib` both as a production dependency (`"^2.0.0"`) and as a peer dependency (`"2"`). As reported in #1241, this causes pnpm to emit an unmet/missing peer dependency warning in consuming projects that don't declare `tslib` themselves, even though memfs already resolves it through its own `dependencies`.

Per the maintainer's guidance on the issue:

> Probably should be just `"^2.0.0"` in `"dependencies"`. Not in `"peerDependencies"`.

## Change

Removes the `peerDependencies` entry for `tslib`, keeping `"tslib": "^2.0.0"` in `dependencies`. Package-manifest-only change; no runtime code is touched.

### DIFF
--- a/packages/memfs/package.json
+++ b/packages/memfs/package.json
@@ -124,9 +124,6 @@
     "webpack-cli": "^5.1.4",
     "webpack-dev-server": "^4.15.1"
   },
-  "peerDependencies": {
-    "tslib": "2"
-  },
   "dependencies": {
     "@jsonjoy.com/fs-core": "workspace:*",
     "@jsonjoy.com/fs-fsa": "workspace:*",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4976,8 +4976,6 @@ __metadata:
     webpack: "npm:^5.87.0"
     webpack-cli: "npm:^5.1.4"
     webpack-dev-server: "npm:^4.15.1"
-  peerDependencies:
-    tslib: 2
   languageName: unknown
   linkType: soft
 


unrevised6419:
IIRC. Make sure to run `yarn install` so lock file to be updated.

xianjianlf2:
Good catch — updated `yarn.lock` to drop the now-removed tslib peer entries (2-line change). I confirmed the remaining `YN0002 ... doesn't provide tslib` warnings for the sibling `@jsonjoy.com/fs-*` workspaces are pre-existing: the count is identical with or without this change, so removing the peerDependency doesn't introduce them.

unrevised6419 (approved):
LGTM

streamich (approved):
Thank you!

PR #1273, opened 2026-07-30
feat: persist git-fsa directory handle

lux-in-tenebris-lucet:
## Summary

Update the browser Git + File System Access demo so it can remember and reopen the selected folder.

This pattern is useful for browser-based Git clients, editors, IDEs, local-first apps, and content tools that work directly with user folders. Those products can resume the same folder after a reload instead of asking users to select it again or accidentally initializing the demo repository twice.

## What changed

- Store the selected directory handle in IndexedDB.
- Restore access automatically when permission remains granted.
- Provide reconnect and forget actions when permission must be restored.
- Reopen an existing demo repository without creating another initial commit.
- Repair the demo's browser build after the monorepo migration and remove its unused worker.
- Document the run command and manual test flow.

## Why

The existing demo was a one-time flow: it forgot the selected folder on reload and always tried to create a new `/repo` directory. Its browser build also referenced an old worker path and did not handle the `node:` imports used by the current workspace packages.

## Testing

- `yarn test` — 87 suites and 1,244 tests passed
- `yarn typecheck`
- `yarn prettier:check`
- `yarn build`
- `yarn workspace memfs webpack --config ./demo/git-fsa/webpack.config.js`
- Manually created the demo repository, reloaded the page, reopened the saved folder, and verified that only one initial commit existed

### DIFF
--- a/packages/memfs/demo/git-fsa/README.md
+++ b/packages/memfs/demo/git-fsa/README.md
@@ -1,18 +1,39 @@
-This demo showcase how to run Git in browser but write to a real user file system
-folder. It is possible through File System Access API. The API allows to request
-a folder from user and then use it as a real file system in browser.
+# Git in a real browser folder
 
-In this demo we use `memfs` to create a Node `fs`-like file system in browser
-out of the folder provided by File System Access API. We then use `isomorphic-git`
-to run Git commands on that file system.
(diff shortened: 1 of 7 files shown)


streamich (approved):
Thank you!

PR #1275, opened 2026-08-03
fix(build): embed sources in published sourcemaps

### LINKED ISSUE #1258: Published JS sourcemaps reference omitted src files
### Summary

The published `memfs` tarball includes JavaScript sourcemaps that point at TypeScript sources which are not included in the package, and the maps do not embed `sourcesContent`. Tools that try to hydrate source content (for example Vite 8) warn when `memfs` is imported.

I confirmed this in `memfs@4.57.1` and the current `memfs@4.57.2` tarball.

### Example

`lib/index.js` ends with:

```js
//# sourceMappingURL=index.js.map
```

`lib/index.js.map` contains:

```json
{
  "sources": ["../src/index.ts"]
}
```

but the npm tarball has no `src/index.ts`, and the map has no `sourcesContent` array. The other published `lib/**/*.js.map` entry maps appear to follow the same pattern.

### User-visible symptom

With Vite, importing `memfs` can emit:

```txt
Sourcemap for ".../node_modules/memfs/lib/index.js" points to missing source files
```

With `DEBUG=vite:sourcemap`, Vite shows the exact missing path:

```txt
Missing sources:
  .../node_modules/memfs/src/index.ts
```

### Possible fixes

Any of these would resolve it:

1. Embed `sourcesContent` in the published `.js.map` files.
2. Publish the referenced `src/**/*.ts` files.
3. If source maps are not intended to be useful to consumers, omit the emitted source map comments/maps from the tarball.

Thanks!

xianjianlf2:
## Summary

Fixes #1258.

TypeScript emits JavaScript sourcemaps with references to `src/**/*.ts`, but the published packages only include `lib`. Because `sourcesContent` was not embedded, tools such as Vite could not hydrate the referenced sources and reported missing-source warnings.

Enable `inlineSources` in the shared TypeScript configuration so all package builds embed their source content in published sourcemaps.

## Validation

- `yarn build`
- Verified all 81 generated `lib/**/*.js.map` files contain `sourcesContent`
- `yarn workspace memfs exec jest --runInBand`
- `yarn prettier:check`

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
     "noImplicitAny": false,
     "strictNullChecks": true,
     "sourceMap": true,
+    "inlineSources": true,
     "outDir": "lib",
     "declaration": true,
     "skipLibCheck": true,


streamich (approved):
Thank you!

PR #1284, opened 2026-09-08
fix(fs-node): match Node directory iteration lifecycle

xianjianlf2:
Bring `Dir` end-of-iteration behavior in line with Node.js: repeated reads after EOF keep returning `null`, and async iteration closes the directory on completion, `break`, or an exception in the consumer. Make explicit resource disposal idempotent so it remains safe after automatic iterator cleanup; direct `close()` still reports `ERR_DIR_CLOSED` on a second close.

The same lifecycle regression tests run against native Node.js and memfs. Before the fix all five native cases pass on Node 24 and all five memfs cases fail. One existing Dirent test was updated to rely on the iterator's automatic close instead of closing the directory a second time.

Validation: full `yarn test`, workspace typechecks, and repository Prettier check. Resource-disposal comparisons are skipped when the test environment does not expose the disposal symbols or the native API. Node 22 full suite: 1,446 passed, 20 skipped. Node 24: all 10 lifecycle comparison cases pass; full suite: 1,448 passed, 18 skipped.

AI-assisted implementation and validation.

### DIFF
--- a/packages/fs-node/src/Dir.ts
+++ b/packages/fs-node/src/Dir.ts
@@ -27,6 +27,7 @@ export class Dir implements IDir {
   }
 
   private readBase(iteratorInfo: IterableIterator<[string, Link | undefined]>[]): IDirent | null {
+    if (iteratorInfo.length === 0) return null;
     let done: boolean | undefined;
     let value: [string, Link | undefined];
     let name: string;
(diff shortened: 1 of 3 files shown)


streamich (approved):
Thank you!

PR #1285, opened 2026-09-08
fix(fs-node): finish BYOB file stream reads at EOF

xianjianlf2:
A BYOB reader of `FileHandle.readableWebStream()` never resolves its final `read()` when the file reaches EOF, including for an empty file. Closing a byte stream does not settle an outstanding BYOB request: respond with zero bytes after closing the controller so the reader receives `done: true`.

The regression suite consumes empty and non-empty files with both default and BYOB readers against native Node.js and memfs. Before the fix both memfs BYOB cases time out; the corresponding native cases and default-reader controls pass.

Validation on Node 24: full `yarn test` passed (1,446 tests, 18 existing skips, 14 snapshots), workspace typechecks and repository Prettier check passed.

AI-assisted implementation and validation.

### DIFF
--- a/packages/fs-node/src/FileHandle.ts
+++ b/packages/fs-node/src/FileHandle.ts
@@ -152,6 +152,7 @@ export class FileHandle extends EventEmitter implements IFileHandle {
 
           if (result.bytesRead === 0) {
             controller.close();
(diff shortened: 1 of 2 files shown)


streamich (approved):
Thank you!

